PR #57, opened 2016-07-15
Fixed Logic For accessSync Use

Hunga1:
Removed conditional if to comply with [accessSync interface](https://nodejs.org/api/fs.html#fs_fs_accesssync_path_mode).


kethinov:
Good catch. Thanks.

PR #58, opened 2016-07-28
Adds HTML Validation to Roosevelt

dishad:
Added a developer dependency for `html-validator`, and added a function call for the HTML validator to `roosevelt.js`.  Includes an error page for visualization of the errors


kethinov:
Couple issues with this:

- Doesn't work with the sample app out of the box for a whole host of reasons. Generate a sample app with `mkroosevelt` and test this PR against that.
- The right place for the `html-minifier` dep is in dependencies, not devDependencies.

kethinov:
Thanks @dishad for this. We cleaned it up and got it merged here, so I'll close this PR. https://github.com/kethinov/roosevelt/pull/89

PR #61, opened 2016-08-19
HTTPS Server Support

jkae:
Adds support for dual and either-or HTTP/HTTPS server creation.

It's a bit messy at the moment, but all the functionality is there. I'll update my fork with any suggestions or necessary changes.

### DIFF
--- a/lib/sourceParams.js
+++ b/lib/sourceParams.js
@@ -44,7 +44,7 @@ module.exports = function(app) {
     pfx: params.pfx || pkg.rooseveltConfig.pfx || false,
     keyPath: params.keyPath || pkg.rooseveltConfig.keyPath, // object with pfx / key+cert (file paths)
     passphrase: params.passphrase || pkg.rooseveltConfig.passphrase, // string
-    ca: params.ca || pkg.rooseveltConfig.ca, // string or array of strings (file paths)
+    ca: params.ca || pkg.rooseveltConfig.ca, // string or array of strings (file paths). if used, server will request authorization & reject unauthorized users
 
     // mvc
     modelsPath: params.modelsPath || pkg.rooseveltConfig.modelsPath || 'mvc/models',
--- a/lib/sourceParams.js
+++ b/lib/sourceParams.js
@@ -37,13 +37,14 @@ module.exports = function(app) {
     multipart: params.multipart || pkg.rooseveltConfig.multipart || {'multiples': true},
     maxLagPerRequest: params.maxLagPerRequest || pkg.maxLagPerRequest || 2000,
     shutdownTimeout: params.shutdownTimeout || pkg.shutdownTimeout || 30000,
-    // https behavior
+    // https behavior - generally no defaults, user-defined
     https: params.https || pkg.rooseveltConfig.https || false,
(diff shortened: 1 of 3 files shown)


roosevelt.js
```diff
@@ -21,16 +23,59 @@ module.exports = function(params) {
...
+      else {
+        httpsOptions.key = fs.readFileSync(app.get('params').keyPath.key);
+        httpsOptions.cert = fs.readFileSync(app.get('params').keyPath.cert);
+      }
+      if (passphrase) {
+        httpsOptions.passphrase = passphrase;
+      }
+      if (ca) {
+        // String or array
+        if (typeof ca === String) {
```
Hunga1: The `typeof` operator outputs a string representing the operand's type. This would need to be the string literal `"string"`

roosevelt.js
```diff
@@ -21,16 +23,59 @@ module.exports = function(params) {
...
+          httpsOptions.ca = fs.readFileSync(ca);
+        }
+        else if (typeof ca === Array) {
```
Hunga1: `typeof` only recognizes primitive types and an `Object`. You can use the `instanceof` operator to compare the variable with the `Array` object to test the variables prototype chain against Array (i.e. `ca instanceof Array`). It outputs a boolean.

roosevelt.js
```diff
@@ -104,6 +156,21 @@ module.exports = function(params) {
       };
   
   function startServer() {
+    var lock = {},
```
Hunga1: I have a question about the lock. Did you implement it so that the `onServerStart` event fired only once in the case that HTTPS was enabled and `httpsOnly` was disabled (two servers)?

jkae: Yes, my assumption was onServerStart should only fire once in this case.

lib/sourceParams.js
```diff
@@ -37,6 +37,16 @@ module.exports = function(app) {
...
     maxLagPerRequest: params.maxLagPerRequest || pkg.maxLagPerRequest || 2000,
     shutdownTimeout: params.shutdownTimeout || pkg.shutdownTimeout || 30000,
+    // https behavior - generally no defaults, user-defined
+    https: params.https || pkg.rooseveltConfig.https || false,
+    httpsOnly: params.httpsOnly || pkg.rooseveltConfig.httpsOnly || false,
+    httpsPort: params.httpsPort || pkg.rooseveltConfig.httpsPort || 43733,
+    pfx: params.pfx || pkg.rooseveltConfig.pfx || false,
+    keyPath: params.keyPath || pkg.rooseveltConfig.keyPath, // object with pfx / key+cert (file paths)
+    passphrase: params.passphrase || pkg.rooseveltConfig.passphrase, // string
+    ca: params.ca || pkg.rooseveltConfig.ca, // string or array of strings (file paths)
```
Hunga1: Could you add an additional OR so that `keyPath`, `passphrase`, and `ca` will be `null` if the `params` or Roosevelt config are undefined. Just to keep it consistent with the defaults specified in your documentation.

roosevelt.js
Hunga1: Great thanks.

PR #63, opened 2016-10-14
limit parameter

matthew-renodin:
Added support for the limit parameter allowing users to set the limit coming from the package.json. It is applied in the the express body parser and in the JSON parser. This limit is necessary when POSTing data greater than 100kb in size.

### DIFF
--- a/lib/setExpressConfigs.js
+++ b/lib/setExpressConfigs.js
@@ -52,9 +52,14 @@ module.exports = function(app) {
 
   // defines req.body by parsing http requests
   app.use(bodyParser.urlencoded({
-    extended: true
+    extended: true,
+    limit: params.limit
   }));
-  app.use(bodyParser.json());
+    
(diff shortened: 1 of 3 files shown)


lib/setExpressConfigs.js
```diff
@@ -52,9 +52,14 @@ module.exports = function(app) {
...
   // defines req.body by parsing http requests
   app.use(bodyParser.urlencoded({
-    extended: true
+    extended: true,
+    limit: params.limit
+  }));
+    
+  // when the HTTP request contains JSON data this parser is used
+  app.use(bodyParser.json({
+      limit: params.limit
```
Hunga1: Fix the indentation to match the conventions used in the rest of the file.

matthew-renodin:
@Hunga1 updated the documentation and fixed the indent.

PR #64, opened 2016-10-28
limit parameter

matthew-renodin:
Added the limit parameter and functionality to the bodyParser with the README update


kethinov:
Better to just allow users to define body-parser's [entire params object](https://github.com/expressjs/body-parser#bodyparserjsonoptions) rather than just `limit`. That would allow you to define other params in addition to `limit`.

kethinov:
Doing that will also close https://github.com/kethinov/roosevelt/issues/50

matthew-renodin:
Great catch. I just wrote an update for this.

PR #72, opened 2016-12-08
Added Travis yml file for Travis builds

alallier:
Please DO NOT merge until #71 is merged

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,3 @@
+language: node_js
+node_js:
+  - "node"


alallier:
This runs the builds against the latest node (see [Travis Available-Versions](https://docs.travis-ci.com/user/languages/javascript-with-nodejs/#Available-Versions)

And `npm test` is ran automatically (which will be handled by #71

alallier:
We could also consider turning on [required status checks](https://help.github.com/articles/enabling-required-status-checks/), which doesn't allow merging unless the build it passed

PR #74, opened 2016-12-09
Fixed pathing issues for all operating systems whether using the whitelist or not (Closes #70)

### LINKED ISSUE #70: Error Compiling Static Javascript Files
When trying to start the application. The following error occurs:
```
Error: ENOENT: no such file or directory, lstat '/Users/username/webapp/statics/js/Users/username/webapp/statics/js/jsFile.js'
```

I'm not using the `jsCompilerWhitelist` configuration. It seems like when Roosevelt attempts to compile all static javascript files, it is appending the absolute path to the statics directory twice. This error seemed to be introduced in the latest version `0.6.30`.

alallier:


lib/jsCompiler.js
```diff
@@ -40,11 +41,13 @@ module.exports = function(app) {
 
   jsFiles.forEach(function(file) {
     (function(file) {
-      var split = file.split(':'),
-          altdest = split[1];
-      file = split[0];
-      if (file === '.' || file === '..' || file === 'Thumbs.db' || fs.lstatSync(path.normalize(app.get('jsPath') + file)).isDirectory()
-) {
+      if (usingWhitelist === true) {
+        var split = file.split(':'),
```
kethinov: better to declare these vars outside of the if statement, then define them inside the if statement.

alallier:
Updated

alallier:
Updated

PR #102, opened 2017-03-09
Make HTML validator behave

Autre31415:


lib/htmlValidator.js
```diff
@@ -55,7 +55,7 @@ module.exports = function(app) {
       // get the html for this specific render
       app.render(view, model, function(err, html) {
 
-        if (process.env.NODE_ENV === 'development' && !validatorDisabled) {
+        if (process.env.NODE_ENV === 'development' && !validatorDisabled && req.header !== 'Partial' && !model._disableValidator) {
```
kethinov: - `req.header` seems wrong 
- these escape hatches need documentation in the README

README.md
```diff
@@ -236,6 +236,7 @@ App behavior parameters
   - Default: `false`
 - `htmlValidator`: Params to send to [html-validator](https://github.com/zrrrzzt/html-validator#usage) (if `enableValidator` is set to true). When `suppressWarnings` is set to true validation warnings will be hidden and only errors will be shown.
   - Default:  `{validator: 'http://html5.validator.nu', format: 'text', suppressWarnings: true}`
+  - Can also be disabled via request by setting a 'Partial' header and by passing `_disableValidator` into the model.
```
kethinov: "Can be disabled for individual requests by sending the request header `Partial` with the value set to `true` or by passing `_disableValidator` to the model and setting it to `true`."

We also need to make settings to customize these params so they can be something other than `Partial` and `_disableValidator`.

README.md
```diff
@@ -236,7 +236,9 @@ App behavior parameters
   - Default: `false`
 - `htmlValidator`: Params to send to [html-validator](https://github.com/zrrrzzt/html-validator#usage) (if `enableValidator` is set to true). When `suppressWarnings` is set to true validation warnings will be hidden and only errors will be shown.
   - Default:  `{validator: 'http://html5.validator.nu', format: 'text', suppressWarnings: true}`
-  - Can also be disabled via request by setting a 'Partial' header and by passing `_disableValidator` into the model.
+  - Can be disabled for individual requests by sending the request header Partial with the value set to true or by passing _disableValidator to the model and setting it to true.
```
kethinov: need to preserve the formatting i supplied in my previous code comment

README.md
```diff
@@ -236,7 +236,9 @@ App behavior parameters
...
+- `validatorExceptions`: Params to customize which request request header or model value can be used to disable the HTML validator.
```
kethinov: "Use this to customize the name of the request header or model value that is used to disable the HTML validator."

PR #116, opened 2017-03-31
Resolve app crashing if moved/renamed (Closes #44)

### LINKED ISSUE #44: App crashes if started, symlinks made, app moved/renamed, then restarted
Should either support this, or throw an error that is less cryptic.

Autre31415:
In order to smooth out app operations after a move we need to properly detect broken symlinks and replace them with new ones:

Enter `checkFiles.js`, the heroic helper module which contains our functions to check both the existence of files and symlinks, allowing us to wipe away any broken symlinks and pave the way for new ones to be made. As a bonus all instances of `fs.existsSync` have been replaced with the already integrated `fileExists` function, and if (more like when) the fs module changes the file detection method we only need to update that one function to cover the whole app.

Also tossed in some OCD cleanup of other modules

### DIFF
--- a/lib/checkFiles.js
+++ b/lib/checkFiles.js
@@ -24,4 +24,4 @@ module.exports = {
       return false;
     }
   }
-}
+};
--- a/lib/checkFiles.js
+++ b/lib/checkFiles.js
@@ -1,0 +1,27 @@
+// check existence of files and symlinks
+
+var fs = require('fs');
+
+module.exports = {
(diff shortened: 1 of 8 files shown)


kethinov:
@alallier please review this if you get a chance. I won't be able to properly test it until Monday at the earliest. If I get to it before you, I'll merge it though.

LGTM with just a cursory glance at the code though.

PR #118, opened 2017-04-13
bring back toobusy (closes #31)

### LINKED ISSUE #31: Restore or replace toobusy
I had to disable toobusy because it was breaking npm installs of fresh roosevelt apps. Seems related to this: https://github.com/lloyd/node-toobusy/issues/45

Autre31415:
toobusy has been broken for a long time now due to its lack of maintenance, but now we have this very handy fork of it which rewrote the feature in JS!

That can be found here: https://github.com/STRML/node-toobusy

The API is exactly the same, so I was able to reuse the old code from the original implementation, I only changed the default maxLag setting to the suggested default of 70.

I tested that we hit the 503 by creating a slow route (Found in their readme) and trying to load the slow route and a normal route at the same time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -242,7 +242,7 @@ App behavior parameters
 - `multipart`: Settings to pass along to [formidable](https://github.com/felixge/node-formidable) using [formidable's API](https://github.com/felixge/node-formidable#api) for multipart form processing. Access files uploaded in your controllers by examining the `req.files` object. Roosevelt will remove any files uploaded to the `uploadDir` when the request ends automatically. To keep any, be sure to move them before the request ends. To disable multipart forms entirely, set this option to false.
   - Default: `{'multiples': true}`
 - `maxLagPerRequest`: Maximum amount of time in miliseconds a given request is allowed to take before being interrupted with a 503 error. (See [node-toobusy](https://github.com/STRML/node-toobusy))
-  - Default: `2000` (2 seconds)
+  - Default: `70` (70ms)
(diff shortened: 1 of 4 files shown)


kethinov:
Need to update the README.

Otherwise LGTM.

README.md
```diff
@@ -241,7 +241,7 @@ App behavior parameters
   - Default: `{'requestHeader': 'Partial', 'modelValue': '_disableValidator'}`
 - `multipart`: Settings to pass along to [formidable](https://github.com/felixge/node-formidable) using [formidable's API](https://github.com/felixge/node-formidable#api) for multipart form processing. Access files uploaded in your controllers by examining the `req.files` object. Roosevelt will remove any files uploaded to the `uploadDir` when the request ends automatically. To keep any, be sure to move them before the request ends. To disable multipart forms entirely, set this option to false.
   - Default: `{'multiples': true}`
-- ~~`maxLagPerRequest`: Maximum amount of time in miliseconds a given request is allowed to take before being interrupted with a 503 error. (See [node-toobusy](https://github.com/lloyd/node-toobusy))~~ *([Temporarily disabled](https://github.com/lloyd/node-toobusy/issues/45))*
+- `maxLagPerRequest`: Maximum amount of time in miliseconds a given request is allowed to take before being interrupted with a 503 error. (See [node-toobusy](https://github.com/STRML/node-toobusy))
```
kethinov: you didn't update the default

PR #148, opened 2017-06-01
prevent files from being set to undefined

Autre31415:
Judging by some digging with console logs I've determined that when these functions are called on server init they are being passed the file arg properly, and then `file = file.path` results in setting them to undefined, which creates a cascade of errors when any manipulation is done to them later on. Removing these lines allows the server to run as normal.

### DIFF
--- a/lib/jsCompiler.js
+++ b/lib/jsCompiler.js
@@ -40,7 +40,6 @@ module.exports = function(app, callback) {
   }
 
   jsFiles.forEach(function(file) {
-    file = file.path;
     promises.push(function(file) {
       return new Promise(function(resolve, reject) {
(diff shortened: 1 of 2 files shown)


kethinov:
This seems overly simplistic. Those lines were added for a very specific reason [here]( https://github.com/rooseveltframework/roosevelt/commit/4545203ea775500fdb8e9edacbaf65f1feb75cdf).

Switching to klaw-sync instead of fse's walkSync changes the API. It returns a file list with `.path` as the actual file path location. Removing those lines might suppress the error, but I'm almost certain it will introduce a new bug, based on this context.

Can you take a deeper look at this?

PR #151, opened 2017-06-01
Better error handling when Java is not installed, fixed html validation issue on Mac

### LINKED ISSUE #146: Provide better error when Java isn't installed (or in path) for HTML validation
Not sure of the error I encounter earlier with someone but the error printed when Java wasn't installed was not helpful. Even if we can't easily determine that specific error it might be nice to just provide a general warning if we encounter an error because most likely Java isn't installed.

Like: Some error (probably Java isn't installed), Warning HTML validation disabled error on initialization (check to sure Java is installed and in your path)

Related to #145

JoshL18:
(closes #146)

Created better error message when Java is not installed. Fixed issue with validation process hanging and displaying no error message on Mac.

### DIFF
--- a/lib/htmlValidator.js
+++ b/lib/htmlValidator.js
@@ -39,7 +39,6 @@ module.exports = function(app, callback) {
       params.htmlValidator.validator = 'http://localhost:' + (params.htmlValidator.port ? params.htmlValidator.port : '8888');
       options = params.htmlValidator;
     }
-
     // spawn validator child process
     promise = (() => {
       return new Promise((resolve, reject) => {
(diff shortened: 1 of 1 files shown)


lib/htmlValidator.js
```diff
@@ -53,9 +52,24 @@ module.exports = function(app, callback) {
...
+          else {
+            resolve();
+          }
+        });
+
         validatorProcess.on('error', (err) => {
-          console.error(`error: ${err}`);
-          reject();
+          if (err.code === 'ENOENT' ) {
+            console.error('error: You must install Java to continue. Warning: HTML validation disabled, error on initialization (check to make sure Java is installed and in your path)');
```
kethinov: this seems kinda redundant. there's a warning and an error saying the same thing twice on one line... pick one?

lib/htmlValidator.js
```diff
@@ -53,9 +52,24 @@ module.exports = function(app, callback) {
           }
         });
 
+        validatorProcess.stderr.on('data', (data) => {
+          if (`${data}`.includes('No Java runtime present, requesting install.')) {
+            console.error('There was an error in authentication. Please make sure to have Java installed.');
```
kethinov: Let's change the wording here to "No Java runtime detected. Check to make sure Java is installed and in your path."

Both this error and the other one down there can probably use the same language.

PR #154, opened 2017-06-05
Timeout on HTML Validator

### LINKED ISSUE #145: Timeout on validator initalization
We should set a good chunk of time before we decide that for some reason html validator initialization failed and then just print an error to the user saying validation is disabled because of a problem and move on. Otherwise the app will hang and never start because right now the validation initialization is blocking; if it never completes the app never finishes starting

Maybe something like: Wait 30 seconds or so, provide the warning, and print standard out from the `exec` (so that they can file an issue with that or look into things further?)

Related to #143 and #135

JoshL18:
(closes #145)

Created a 10 second timeout on the HTML Validator to ensure whether the validation completed. Initialization continues even in the case that the validator has been disabled. Errors during validation initialization are now printed to console in red for better visibility.

### DIFF
--- a/lib/htmlValidator.js
+++ b/lib/htmlValidator.js
@@ -49,7 +49,7 @@ module.exports = function(app, callback) {
           'java', ['-Xss1024k', '-cp', vnu, 'nu.validator.servlet.Main', params.htmlValidator.port ? params.htmlValidator.port : '8888']
         );
 
-        validatorProcess.stdout.on('data', (data) => { 
+        validatorProcess.stdout.on('data', (data) => {
           if (`${data}`.includes('Initialization complete')) {
             validatorDisabled = false;
             clearTimeout(validatorTimeout);
@@ -171,10 +171,10 @@ module.exports = function(app, callback) {
     };
     next();
   });
-  
-  promise.then(function(reason){
+
+  promise.then(function (reason) {
     callback();
-  }).catch(function(reason) {
+  }).catch(function (reason) {
     // rejection
     console.error(('HTML Validator disabled. Continuing without HTML validation...').red);
     callback();
--- a/lib/htmlValidator.js
+++ b/lib/htmlValidator.js
@@ -26,7 +26,7 @@ module.exports = function(app, callback) {
       formattedHTML,
       validatorTimeout,
       errorPage = path.normalize(__dirname + '/../defaultErrorPages/views/validatorErrors'),
-      validatorDisabled = true;
+      validatorDisabled;
 
   if (!params.enableValidator) {
     validatorDisabled = true;
@@ -49,10 +49,12 @@ module.exports = function(app, callback) {
           'java', ['-Xss1024k', '-cp', vnu, 'nu.validator.servlet.Main', params.htmlValidator.port ? params.htmlValidator.port : '8888']
         );
 
+        console.log(('Validator intializing...').yellow);
+
         validatorProcess.stdout.on('data', (data) => {
           if (`${data}`.includes('Initialization complete')) {
-            validatorDisabled = false;
             clearTimeout(validatorTimeout);
+            console.log(('Validator initialization completed successfully.').green);
             resolve();
           }
         });
@@ -68,6 +70,7 @@ module.exports = function(app, callback) {
         validatorProcess.on('error', (err) => {
           if (err.code === 'ENOENT' ) {
             console.error(('You must install Java to continue. HTML validation disabled, error on initialization (check to make sure Java is installed and in your path)').red);
+            validatorDisabled = true;
             reject();
           }
           else {
@@ -78,11 +81,10 @@ module.exports = function(app, callback) {
         });
 
         validatorTimeout = setTimeout(function() {
-          if (validatorDisabled) {
-            console.error(('HTML Validator has timed out.').red);
-            reject();
-          }
-        }, 10000);
+          console.error(('HTML Validator has timed out.').red);
+          validatorDisabled = true;
+          reject();
+        }, 20000);
       });
     })();
   };
@@ -172,9 +174,9 @@ module.exports = function(app, callback) {
     next();
   });
 
-  promise.then(function (reason) {
+  promise.then(function(result) {
     callback();
-  }).catch(function (reason) {
+  }).catch(function(result) {
     // rejection
     console.error(('HTML Validator disabled. Continuing without HTML validation...').red);
     callback();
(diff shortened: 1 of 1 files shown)


lib/htmlValidator.js
```diff
@@ -22,8 +24,9 @@ module.exports = function(app, callback) {
       htmlArrLength,
       htmlLine,
       formattedHTML,
+      validatorTimeout,
       errorPage = path.normalize(__dirname + '/../defaultErrorPages/views/validatorErrors'),
-      validatorDisabled;
+      validatorDisabled = true;
```
Autre31415: This shouldn't need to be hardcoded. It's taken care of in the following if statement.

lib/htmlValidator.js
```diff
@@ -48,29 +51,38 @@ module.exports = function(app, callback) {
 
         validatorProcess.stdout.on('data', (data) => {
           if (`${data}`.includes('Initialization complete')) {
+            validatorDisabled = false;
```
Autre31415: This line becomes redundant when you remove the hardcoding of this variable.

lib/htmlValidator.js
```diff
@@ -48,29 +51,38 @@ module.exports = function(app, callback) {
...
             reject();
           }
         });
+
+        validatorTimeout = setTimeout(function() {
+          if (validatorDisabled) {
+            console.error(('HTML Validator has timed out.').red);
+            reject();
+          }
+        }, 10000);
```
Autre31415: Are we sure that 10 seconds is enough for some slower machines or laggy vms?

lib/htmlValidator.js
```diff
@@ -160,7 +172,11 @@ module.exports = function(app, callback) {
     next();
   });
 
-  Promise.all([promise]).then(() => {
+  promise.then(function (reason) {
```
Autre31415: Remove the space between the function and arg, and rename arg to result

lib/htmlValidator.js
```diff
@@ -160,7 +172,11 @@ module.exports = function(app, callback) {
...
+    callback();
+  }).catch(function (reason) {
```
Autre31415: Same here

lib/htmlValidator.js
```diff
@@ -48,29 +51,38 @@ module.exports = function(app, callback) {
...
         });
 
         validatorProcess.stderr.on('data', (data) => {
           if (`${data}`.includes('No Java runtime present, requesting install.')) {
-            console.error('error: No Java runtime detected. Check to make sure Java is installed and in your path.');
+            console.error(('There was an error in authentication. Please make sure to have Java installed.').red);
+            validatorDisabled = true;
             reject();
           }
-          else {
```
alallier: Don't mean to step on your toes @Autre31415, I have just been helping @JoshL18 in person with this. You have to be careful with the removal of this else, (is there a case when we will get something in standard error and the validator can't start? If so we will want this else.

I may have told you to remove this @JoshL18, if I did sorry about that.

Autre31415: As far as I'm aware anything we're getting in standard error should cause the promise to be rejected. We're only resolving the promise when we know for sure that the validator is working as intended.

Would be a good idea to test this though. A console dump of all data under stderr should clarify.

Autre31415: As far as I'm aware anything we're getting in standard error should cause the promise to be rejected. We're only resolving the promise when we know for sure that the validator is working as intended.

Would be a good idea to test this though. A console dump of all data under stderr should clarify.

alallier: Could there be a race condition, if we get some standard error, but everything still initiates, resulting in a reject and then a resolve?

Autre31415: Once a promise is handled it's done. It won't be handled more than once.

lib/htmlValidator.js
alallier: Btw @Autre31415 `reason` seems to be the standard according to [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/catch) not sure if that matters as far as readability/styling is concerned?

lib/htmlValidator.js
Autre31415: Alex and I have discussed this and we're in agreement this else should be left out.

lib/htmlValidator.js
```diff
@@ -46,31 +49,43 @@ module.exports = function(app, callback) {
           'java', ['-Xss1024k', '-cp', vnu, 'nu.validator.servlet.Main', params.htmlValidator.port ? params.htmlValidator.port : '8888']
         );
 
+        console.log(('Validator intializing...').yellow);
```
JoshL18: @kethinov Is the wording here okay?

lib/htmlValidator.js
```diff
@@ -46,31 +49,43 @@ module.exports = function(app, callback) {
...
+
         validatorProcess.stdout.on('data', (data) => {
           if (`${data}`.includes('Initialization complete')) {
+            clearTimeout(validatorTimeout);
+            console.log(('Validator initialization completed successfully.').green);
```
JoshL18: @Autre31415 I added a line to tell the user when initialization was successful, should we keep this or just get rid of it?

lib/htmlValidator.js
kethinov: Go with "Starting validator..."

kethinov: Actually make that "Starting HTML validator..."

lib/htmlValidator.js
kethinov: And "HTML validator listening on port [port]"

use bold for both messages

kethinov: and yeah we only need one message indicating that the validator was started, so pick the line of code that makes the most sense for it to log it and use that message

lib/htmlValidator.js
```diff
@@ -46,31 +49,43 @@ module.exports = function(app, callback) {
...
             resolve();
           }
         });
 
         validatorProcess.stderr.on('data', (data) => {
           if (`${data}`.includes('No Java runtime present, requesting install.')) {
-            console.error('error: No Java runtime detected. Check to make sure Java is installed and in your path.');
+            console.error(('There was an error in authentication. Please make sure to have Java installed.').red);
+            validatorDisabled = true;
```
Autre31415: Instead of setting this to true on each error, you could set it to true in the promise rejection catch.

kethinov:
Emojis! 🎉🎊🎉🎊🎉🎊

PR #157, opened 2017-06-07
Fixed graceful shutdown not executing properly on Mac

### LINKED ISSUE #149: Audit server graceful shutdown
It seems that Roosevelt never shuts down everything on a graceful shutdown and has to fall back to a forced shutdown (*in Mac, not test in Linux, and fine in Windows*). While developing, this can cause Mac users to have and sit and wait for the forced shutdown (at default of 30 seconds).

*Note: Someone could adjust the shutdown timeout to a lower value but it would seem that the graceful shutdown isn't working properly and it should be audited*

> @kethinov I suspect it's related to the multithreading / cluster / child process API changes that node may have made over the past several versions...

JoshL18:
(closes #149)

Fixed issue where graceful shutdown was not executing on Mac causing a long timeout before Roosevelt is finally killed. This is because there were open connections not being closed when Roosevelt receives the kill signal. 

- Added section of code that recognizes when the server establishes a socket, adding it to an array of sockets that Roosevelt has created.
- When Roosevelt receives the kill signal, Roosevelt now closes all open sockets, allowing for the graceful shutdown process to continue as intended on Mac.

### DIFF
--- a/roosevelt.js
+++ b/roosevelt.js
@@ -31,7 +31,9 @@ module.exports = function(params) {
       passphrase,
       numCPUs = 1,
       servers = [],
-      i;
+      i,
+      sockets = {},
+      nextSocketId = 0;
 
   // expose initial vars
   app.set('express', express);
@@ -84,6 +86,17 @@ module.exports = function(params) {
 
   app.httpServer = httpServer;
   app.httpsServer = httpsServer;
+
+  // add open sockets to array when connected
+  httpServer.on('connection', function (socket) {
+    var socketId = nextSocketId++;
+    sockets[socketId] = socket;
+
+    // once the socket closes, remove socket from array
+    socket.once('close', function () {
+      delete sockets[socketId];
+    });
+  });
 
   // enable gzip compression
   app.use(require('compression')());
@@ -179,6 +192,12 @@ module.exports = function(params) {
             exitLog();
           }
         });
+
+        // destroy sockets when server is killed
+        for (var socketId in sockets) {
+          sockets[socketId].destroy();
+        }
+
         setTimeout(function() {
           console.error(('💥  ' + (app.get('appName') || 'Roosevelt') + ' could not close all connections in time; forcefully shutting down.').red);
           process.exit(1);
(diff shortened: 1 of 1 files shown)


roosevelt.js
```diff
@@ -179,6 +192,12 @@ module.exports = function(params) {
             exitLog();
           }
         });
+
+        // destroy sockets when server is killed
+        for (var socketId in sockets) {
```
kethinov: add socketId to the var list above as well. don't declare it here

alallier:
@kethinov Could this solution cause any memory problems on scaling and what not since we are saving all opened connections in memory?

roosevelt.js
```diff
@@ -85,6 +87,17 @@ module.exports = function(params) {
   app.httpServer = httpServer;
   app.httpsServer = httpsServer;
 
+  // add open sockets to array when connected
+  httpServer.on('connection', function (socket) {
+    var socketId = nextSocketId++;
```
alallier: Do we need the extra var? Could we just make the below line be `sockets[nextSocketId++] = socket;` Since it is defined above across the whole file?

kethinov:
is it possible we could flush the cache over time?

alallier:
@kethinov I was thinking something to that extent I _think_ what we could probably do it when the connection closes also delete it from the cache so that it won't bloat over time. So that way at any given time the cache only holds the open connections, which is exactly what we needed anyways

kethinov:
isn't that what [this](https://github.com/rooseveltframework/roosevelt/pull/157/files#diff-61fd2d5b58856e4efc30ebfbe524bd82R97) does?

do we need to flush it elsewhere?

alallier:
Ah yes you're right it does flush it, we should be good. What about a large number of open connections any problem there?

kethinov:
the whole concept of tracking it in array is kinda gross

i'd rather track it as an object with unique keys rather than a constantly incrementing integer

try this approach? https://github.com/isaacs/server-destroy/blob/master/index.js

alallier:
It is in an object

kethinov:
based on a constantly incrementing integer

look at the alternative solution. it derives the key from the connection itself, delegating the uniqueness to the api rather than us constantly incrementing an integer to create a unique key

alallier:
Understand what you are saying now, thanks for the clarification.

PR #158, opened 2017-06-08
Fixed error with promise when html validator is disabled

### LINKED ISSUE #156: When validator is disabled it throws error because the promise is encapsulated
When validator is disabled it throws error because promise is encapsulated. Meaning the promise is only defined when the param for validation is set to true, so when we reach the [promise resolution code](https://github.com/rooseveltframework/roosevelt/blob/master/lib/htmlValidator.js#L173-L180) it errors because the promise doesn't exist when validation is disabled.

Two ways to solve this:

1. Wrap the promise resoultion at the bottom in the similar if statement found above
2. Go back to promise all (because it can accept an empty array), and give actual promise try catch handling before hand

JoshL18:
(closes #156 )

Moved back to resolving the promise using Promise.all. This change:
- Fixes an error that occurs when the HTML validator has been turned off prior to launching the application. 

- In the case that the HTML validator was turned off, the user is now told that the HTML validator is disabled and to check to make sure that their HTML validator is turned on.

### DIFF
--- a/lib/htmlValidator.js
+++ b/lib/htmlValidator.js
@@ -169,11 +169,10 @@ module.exports = function(app, callback) {
     };
     next();
   });
-
-  promise.then(function(result) {
+  
+  Promise.all([promise]).then((result) => {
     callback();
-  }).catch(function(result) {
-    // rejection
+  }).catch((result) => {
     console.error(('HTML Validator disabled. Continuing without HTML validation...').red);
     validatorDisabled = true;
     callback();
--- a/lib/htmlValidator.js
+++ b/lib/htmlValidator.js
@@ -43,50 +43,47 @@ module.exports = function(app, callback) {
       options = params.htmlValidator;
     }
 
-    // only run validator if in development mode
-    if (process.env.NODE_ENV === 'development') {
-      // spawn validator child process
-      promise = (() => {
-        return new Promise((resolve, reject) => {
-          validatorProcess = spawn(
-            'java', ['-Xss1024k', '-cp', vnu, 'nu.validator.servlet.Main', params.htmlValidator.port ? params.htmlValidator.port : '8888']
-          );
+    // spawn validator child process
+    promise = (() => {
+      return new Promise((resolve, reject) => {
+        validatorProcess = spawn(
+          'java', ['-Xss1024k', '-cp', vnu, 'nu.validator.servlet.Main', params.htmlValidator.port ? params.htmlValidator.port : '8888']
+        );
 
-          console.log('\u23F3  ' + ('Starting HTML validator...').bold.yellow);
+        console.log('\u23F3  ' + ('Starting HTML validator...').bold.yellow);
 
-          validatorProcess.stdout.on('data', (data) => {
-            if (`${data}`.includes('Initialization complete')) {
-              clearTimeout(validatorTimeout);
-              console.log('✔️  ' + ('Validator initialization completed successfully. HTML validator listening on port: ' + params.htmlValidator.port).bold.green);
-              resolve();
-            }
-          });
-
-          validatorProcess.stderr.on('data', (data) => {
-            if (`${data}`.includes('No Java runtime present, requesting install.')) {
-              console.error('❌  ' + ('There was an error in authentication. Please make sure to have Java installed.').red);
-              reject();
-            }
-          });
-
-          validatorProcess.on('error', (err) => {
-            if (err.code === 'ENOENT' ) {
-              console.error('❌  ' + ('You must install Java to continue. HTML validation disabled, error on initialization (check to make sure Java is installed and in your path)').red);
-              reject();
-            }
-            else {
-              console.error('❌  ' + (err).red);
-              reject();
-            }
-          });
-
-          validatorTimeout = setTimeout(function() {
-            console.error('❌  ' + ('HTML Validator has timed out.').red);
-            reject();
-          }, 20000);
+        validatorProcess.stdout.on('data', (data) => {
+          if (`${data}`.includes('Initialization complete')) {
+            clearTimeout(validatorTimeout);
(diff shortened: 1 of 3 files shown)


lib/htmlValidator.js
```diff
@@ -170,10 +170,12 @@ module.exports = function(app, callback) {
     next();
   });
 
-  promise.then(function(result) {
+  Promise.all([promise]).then((result) => {
+    if (!params.enableValidator) {
+      console.error(('HTML Validator disabled. Check to make sure you have HTML Validation enabled. Continuing without HTML validation...').red);
```
JoshL18: @kethinov Is the wording here okay?

kethinov: do a quick audit of all the logs here and change the capitalization from "HTML Validator" to "HTML validator"

as to this message specifically, go with: "⚠️  HTML validator disabled. Continuing without HTML validation..."

also change it to a warning (in yellow) and do not display the warning at all unless it is dev mode (the validation should always be disabled in prod mode)

lib/htmlValidator.js
```diff
@@ -170,10 +170,12 @@ module.exports = function(app, callback) {
...
+    }
     callback();
-  }).catch(function(result) {
-    // rejection
+  }).catch((result) => {
     console.error(('HTML Validator disabled. Continuing without HTML validation...').red);
```
kethinov: ditto above comment here

lib/htmlValidator.js
```diff
@@ -42,47 +42,51 @@ module.exports = function(app, callback) {
...
-        });
 
-        validatorProcess.stderr.on('data', (data) => {
-          if (`${data}`.includes('No Java runtime present, requesting install.')) {
-            console.error('❌  ' + ('There was an error in authentication. Please make sure to have Java installed.').red);
-            reject();
-          }
-        });
+    // only run validator if in development mode
+    if (process.env.NODE_ENV === 'development') {
```
JoshL18: @kethinov checking for dev mode like this is working the way we want it to. Wrapping the promise and validatorProcess in this if check causes the validator to only run when the app is in dev mode. This is now what we wanted it to do, correct?

kethinov: the proper check for this i think is higher up the stack

much simpler to strip all checks for dev mode from this file, then in sourceParams to override the setting

so, set enableValidator to false here: https://github.com/rooseveltframework/roosevelt/blob/5c3beab359378041a2795c85648bab5fe64f87eb/lib/sourceParams.js#L173

then remove all checks for dev mode from htmlValidator.js

lib/htmlValidator.js
```diff
@@ -42,47 +42,51 @@ module.exports = function(app, callback) {
...
+          validatorProcess = spawn(
+            'java', ['-Xss1024k', '-cp', vnu, 'nu.validator.servlet.Main', params.htmlValidator.port ? params.htmlValidator.port : '8888']
+          );
+
+          console.log('\u23F3  ' + ('Starting HTML validator...').bold.yellow);
+
+          validatorProcess.stdout.on('data', (data) => {
+            if (`${data}`.includes('Initialization complete')) {
+              clearTimeout(validatorTimeout);
+              console.log('✔️  ' + ('Validator initialization completed successfully. HTML validator listening on port: ' + params.htmlValidator.port).bold.green);
```
kethinov: change this to:

```js
console.log('🎧  ' + ('HTML validator listening on port: ' + params.htmlValidator.port).bold);
```

lib/htmlValidator.js
kethinov: also move this block of code: https://github.com/rooseveltframework/roosevelt/blob/7fcbe5431749bd0d2988bb7148cd681dd73254b7/roosevelt.js#L120-L127

to just above this: https://github.com/rooseveltframework/roosevelt/blob/7fcbe5431749bd0d2988bb7148cd681dd73254b7/roosevelt.js#L141

lib/sourceParams.js
```diff
@@ -54,10 +57,15 @@ module.exports = function(app) {
 
     // mvc
     modelsPath: params.modelsPath || pkg.rooseveltConfig.modelsPath || 'mvc/models',
+    modelsNodeModulesSymlink: params.modelsNodeModulesSymlink || pkg.rooseveltConfig.modelsNodeModulesSymlink || 'models',
```
kethinov: this param was explicitly removed

lib/sourceParams.js
```diff
@@ -54,10 +57,15 @@ module.exports = function(app) {
...
     viewsPath: params.viewsPath || pkg.rooseveltConfig.viewsPath || 'mvc/views',
     viewEngine: params.viewEngine || pkg.rooseveltConfig.viewEngine || 'html: teddy',
     controllersPath: params.controllersPath || pkg.rooseveltConfig.controllersPath || 'mvc/controllers',
 
+    // utility lib
+    libPath: params.libPath || pkg.rooseveltConfig.libPath || 'lib',
+    libPathNodeModulesSymlink: params.libPathNodeModulesSymlink || pkg.rooseveltConfig.libPathNodeModulesSymlink || 'lib',
```
kethinov: so were these

what happened with this merge?

lib/sourceParams.js
```diff
@@ -116,8 +124,18 @@ module.exports = function(app) {
...
 
+  // explicit falseness check for modelsNodeModulesSymlink
+  if (params.modelsNodeModulesSymlink === false || pkg.rooseveltConfig.modelsNodeModulesSymlink === false) {
+    params.modelsNodeModulesSymlink = false;
+  }
+
+  // explicit falseness check for libPathNodeModulesSymlink
+  if (params.libPathNodeModulesSymlink === false || pkg.rooseveltConfig.libPathNodeModulesSymlink === false) {
+    params.libPathNodeModulesSymlink = false;
+  }
```
kethinov: man you brought back a lot of code that shouldn't be here...

PR #159, opened 2017-06-09
Separated Server Initialization From HTTP Server Startup

### LINKED ISSUE #131: Suggestion to Return Route Binding to Roosevelt Constructor
Prior to Roosevelt v0.7.0, the application's routes were bound by the controller modules when Roosevelt's exported function was executed. After upgrades introduced in v0.7.0 and to the present release, the route binding was moved to the `startServer` method, coupling the assignment of route handlers with the start of the build-in Node HTTP server.

I would like to suggest that route binding be moved back into the Roosevelt constructor/factory method or encapsulated into it's own separate public method. This would benefit Roosevelt by:

1. Decoupling the use of the built-in HTTP server with initializing the Roosevelt application and giving the `startServer` method better separation of concerns
2. Allowing Roosevelt to be used as third party middleware, allowing the framework to be more composable in different architectures.

References:
- A link to the last state which `roosevelt.js` had route binding in the constructor (v0.6.33): https://github.com/kethinov/roosevelt/blob/a45f60d64b0df6d1054fea98532131cf82a5f190/roosevelt.js#L115
- The commit which moved route binding to the `startServer` method: https://github.com/kethinov/roosevelt/commit/963259d9233e3e5685be12b9d112cd73ddd301bf

Hunga1:
Attaching Roosevelt routing and third-party middleware, and compiling of static css, js now happens in a separate public `initServer` method. The `startServer` method calls the `initServer` method if it has not been explicitly called prior. The `initServer` method logic only executes once per Roosevelt object instance, despite repeated calls to the method.

Closes #131

### DIFF
--- a/roosevelt.js
+++ b/roosevelt.js
@@ -31,7 +31,8 @@ module.exports = function(params) {
       passphrase,
       numCPUs = 1,
       servers = [],
-      i;
+      i,
+      initialized = false;
 
   // expose initial vars
   app.set('express', express);
@@ -109,7 +110,12 @@ module.exports = function(params) {
     params.onServerInit(app);
   }
 
-  function startServer() {
+  // Initialize Roosevelt app middleware and prepare static css,js
+  function initServer(cb) {
+    if (initialized) {
+      return cb();
+    }
+    initialized = true;
 
     // initialize HTML validator
     validateHTML();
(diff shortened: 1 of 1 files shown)


kethinov:
Thanks, nice work.

PR #164, opened 2017-06-12
Fixed error with promise when html validator is disabled

### LINKED ISSUE #156: When validator is disabled it throws error because the promise is encapsulated

JoshL18:
(closes #156)

**New pull request due to corruption of the last one.**

 Moved back to resolving the promise using Promise.all. This change:

- Fixes an error that occurs when the HTML validator has been turned off prior to launching the application.
- In the case that the HTML validator was turned off, the user is now told that the HTML validator is disabled and to check to make sure that their HTML validator is turned on.
- Makes it so that the HTML validator does not run when in production mode
- No longer checks for production mode within the HTML validator code

### DIFF
--- a/lib/htmlValidator.js
+++ b/lib/htmlValidator.js
@@ -171,16 +171,16 @@ module.exports = function(app, callback) {
   });
 
   // only resolve promise when in dev mode
-  if (app.get('envStateDev')) {
+  if (process.env.NODE_ENV === 'development') {
     Promise.all([promise]).then((result) => {
       // only print error if validator is disabled and in dev mode
       if (!params.enableValidator) {
-        console.error(('⚠️  ' + 'HTML validator disabled. Continuing without HTML validation...').yellow);
+        console.warn(('⚠️  ' + 'HTML validator disabled. Continuing without HTML validation...').yellow);
       }
       callback();
     }).catch((result) => {
       // only print error if in dev mode
-      console.error(('⚠️  ' + 'HTML validator disabled. Continuing without HTML validation...').yellow);
+      console.warn(('⚠️  ' + 'HTML validator disabled. Continuing without HTML validation...').yellow);
       validatorDisabled = true;
       callback();
     });
--- a/lib/htmlValidator.js
+++ b/lib/htmlValidator.js
@@ -54,7 +54,7 @@ module.exports = function(app, callback) {
         validatorProcess.stdout.on('data', (data) => {
           if (`${data}`.includes('Initialization complete')) {
             clearTimeout(validatorTimeout);
-            console.log('✔️  ' + ('Validator initialization completed successfully. HTML validator listening on port: ' + params.htmlValidator.port).bold.green);
+            console.log('🎧  ' + ('HTML validator listening on port: ' + params.htmlValidator.port).bold);
             resolve();
           }
         });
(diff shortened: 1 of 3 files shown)


lib/sourceParams.js
```diff
@@ -171,6 +171,7 @@ module.exports = function(app) {
 
   if (process.env.NODE_ENV === 'development') {
     params.noMinify = true;
+    params.enableValidator = false;
```
kethinov: this is wrong

the check here is if prod mode enableValidator should be false

kethinov: you will need a separate if statement

roosevelt.js
```diff
@@ -138,6 +129,15 @@ module.exports = function(params) {
       require('./lib/jsCompiler')(app, mapRoutes);
     }
 
+     // initialize HTML validator
+    validateHTML();
+
+    function validateHTML() {
+      require('./lib/htmlValidator')(app, preprocessCss);
```
Autre31415: In order for this sequence of events to be changed `validateHTML();` needs to be removed and replaced with `preProcessCss();`, moved up to line 120 and the callbacks in each function altered to represent the changed sequence.  (With `compileJs` calling `validateHTML` and `validateHTML` calling `mapRoutes`, etc)

kethinov: yeah - just to be super clear: the goal here is to get the css/js parsing stuff to happen first, *then* begin starting the HTML validator so we change the order of the console output

lib/sourceParams.js
```diff
@@ -137,6 +137,16 @@ module.exports = function(app) {
   app.set('cssCompiledOutput', path.normalize(appDir + params.cssCompiledOutput));
   app.set('jsCompiledOutput', path.normalize(appDir + params.jsCompiledOutput));
 
+  // keep track of state of env
+  if (process.env.NODE_ENV === 'development') {
+    // set the envStateDev to true if in dev mode
+    app.set('envStateDev', true);
```
kethinov: we don't need to track this in an app var. querying `process.env.NODE_ENV` each time this is needed is fine

lib/htmlValidator.js
```diff
@@ -170,12 +170,23 @@ module.exports = function(app, callback) {
...
-  }).catch(function(result) {
-    // rejection
-    console.error(('HTML Validator disabled. Continuing without HTML validation...').red);
-    validatorDisabled = true;
+  // only resolve promise when in dev mode
+  if (app.get('envStateDev')) {
+    Promise.all([promise]).then((result) => {
+      // only print error if validator is disabled and in dev mode
+      if (!params.enableValidator) {
+        console.error(('⚠️  ' + 'HTML validator disabled. Continuing without HTML validation...').yellow);
```
kethinov: this should be a warning

lib/htmlValidator.js
```diff
@@ -170,12 +170,23 @@ module.exports = function(app, callback) {
...
+      }
+      callback();
+    }).catch((result) => {
+      // only print error if in dev mode
+      console.error(('⚠️  ' + 'HTML validator disabled. Continuing without HTML validation...').yellow);
```
kethinov: ditto here

Issue #162, opened 2017-06-12
Possible enhancement: Read Roosevelt defaults from (json) file


alallier:
Example instead of:

```javascript
shutdownTimeout: params.shutdownTimeout || pkg.shutdownTimeout || 30000
```

It would become

```javascript
shutdownTimeout: params.shutdownTimeout || pkg.shutdownTimeout || defaultJsonFile.shutdownTimeout
```

This might make the defaults easier to manage, as well as provide an easy place for users to check and for us to refer to on defaults.

kethinov:
Specifically what we're referring to here is a file dedicated to defining the defaults and sourcing that file in sourceParams as well as the yo generator.

alallier:
Just noting that the generator has a [defaults file](https://github.com/rooseveltframework/generator-roosevelt/blob/dfbfc2814e656a9570d25171da8860487ef0cf7c/rooseveltDefaults.json) already, but it contains specific data useful to the generator might be able to join the files.

dishad:
Going to tackle this by creating and importing a JSON file in `lib`

PR #183, opened 2017-06-28
Adding default configuration and fixing parameter priority (Closes #168 and #162)

### LINKED ISSUE #168: Parameter priority
Right now we have three ways for a Roosevelt [parameter](https://github.com/rooseveltframework/roosevelt#configure-your-app-with-parameters) to be populated:

1. By constructor
2. By package.json
3. By default

The way the [source params](https://github.com/rooseveltframework/roosevelt/blob/000fb668b20e8ea7ffab255fad91e68f8313e471/lib/sourceParams.js) file is designed is to give preference by order of above, first check the constructor, if not in the constructor use the `package.json` and lastly if in neither use the default. This is done by using `||` (or) operators.

Example of enableValidator:

```javascript
enableValidator: params.enableValidator || pkg.rooseveltConfig.enableValidator || false 
```

There is a problem with this logic though. And that is in the case of having a constructor param of `false`, and a `package.json` value of `true`. Using the example from above if I have enableValidtor set to `true` in my pacakge.json but specify `false` in a constructor param, the validator will _always_ be set to true because of the `or` statement. The one line `or` check has the param always source from the `package.json` because of the `false` value and `or` logic.

The validator is set to `false` by default in sample apps but when someone goes to set another param to `false` by constructor but the sample app `package.json` has the value set to `true` this will cause unexpected behavior and results.

_Side note: @kethinov I feel you and I have personally discussed this issue before. I don't think I ever filed an issue for this though. I dug through the issues and didn't find anything in there. If this is a duplicate in any way feel free to close and link_

dishad:
Uses `defaultConfig.json` as the configuration file and creates priority for some values that would default to true based on:
1. constructor (most priority)
2. package.json
3. default (least priority)


kethinov:
@dishad throw in the solve for this as well https://github.com/rooseveltframework/roosevelt/issues/168

kethinov:
oh wait you did, i skimmed the diff too fast

kethinov:
it looks like the `!== undefined` approach isn't being applied consistently across the params, which could account for [josh's comment](https://github.com/rooseveltframework/roosevelt/issues/168#issuecomment-312087813)

lib/sourceParams.js
```diff
@@ -28,64 +30,64 @@ module.exports = function(app) {
...
+    https: params.https !== undefined ? params.https : pkg.rooseveltConfig.https !== undefined ? pkg.rooseveltConfig.https : defaultConfig.https,
+    httpsOnly: params.httpsOnly !== undefined ? params.httpsOnly : pkg.rooseveltConfig.httpsOnly !== undefined ? pkg.rooseveltConfig.httpsOnly : defaultConfig.httpsOnly,
+    httpsPort: process.env.HTTPS_PORT || params.httpsPort || pkg.rooseveltConfig.httpsPort || defaultConfig.httpsPort,
+    pfx: params.pfx !== undefined ? params.pfx : pkg.rooseveltConfig.pfx !== undefined ? pkg.rooseveltConfig.pfx : defaultConfig.pfx,
+    keyPath: params.keyPath || pkg.rooseveltConfig.keyPath || defaultConfig.keyPath, // object with pfx / key+cert (file paths)
+    passphrase: params.passphrase || pkg.rooseveltConfig.passphrase || defaultConfig.passphrase, // string
+    ca: params.ca || pkg.rooseveltConfig.ca || defaultConfig.ca, // string or array of strings (file paths)
+    requestCert: params.requestCert !== undefined ? params.requestCert : pkg.rooseveltConfig.requestCert !== undefined ? pks.rooseveltConfig.requestCert : defaultConfig.requestCert,
```
Autre31415: fix this `pks` typo

PR #192, opened 2017-07-08
requirable controllers

kethinov:


lib/mapRoutes.js
```diff
@@ -110,13 +109,18 @@ module.exports = function(app) {
...
 
-    if (error404 < 0) {
+    if (controllerName !== params.error404) {
       try {
         if (fs.statSync(controllerName).isFile()) {
-          require(controllerName)(app);
+          controller = require(controllerName);
+
+          // if the controller accepts more or less than one argument, it's not defining a route
+          if (controller.length === 1) {
```
alallier: What argument is this checking for?

kethinov: If the controller accepts more than one or less than one arguments, then it isn't a route and is therefore presumed to be a requireable controller callback (the callback for a route rather than the route definition + the callback).

I plan to document the feature later to explain this with code examples.

PR #193, opened 2017-07-10
Adding reload (Closes #32)

### LINKED ISSUE #32: Browser auto-reload dev mode feature
Probably using something like this: https://www.npmjs.com/package/reload

alallier:
Some thoughts and questions **before** we merge this:

* Port Stuff
  * Reload automatically runs on port `9856`, if not port is specified. 
  * Do we want to specify our own default port? 
  * If not should we specify the reload default port anyways as people won't know it runs on that port without looking at reload. 
* README Stuff
  * We may need to document how to use reload. As reload is built into Roosevelt you still need to provide the script tag on every page. We could do this when generating the app, but we have to notify people that this is needed somewhere. 
  * Should reload have it's own little section in the README?
* Module Stuff
  * I put reload in it's own module.
  * Do we want this?
    * If so is the name okay?

Other notes:

* Unlike other roosevelt config options I did not create two separate config options to control reload. Right now we have usually have one config to enable/disable something and the other config option for all the options. I personally think it is cleaner. If we want to change this let me know

(Closes #32)


lib/reload.js
```diff
@@ -0,0 +1,7 @@
+module.exports = function(app) {
```
alallier: If we are keeping this module, it should have `//reload` at the top as a comment to keep this consistent with the rest of roosevelt

README.md
```diff
@@ -238,6 +239,8 @@ App behavior parameters
   - Default: `{extended: true}`
 - `bodyParserJsonParams`: Supply parameters to [body-parser.json](https://github.com/expressjs/body-parser#bodyparserjsonoptions).
   - Default: `{}`
+- `reload`: Setup [reload](https://www.npmjs.com/package/reload). Takes an object containing `enabled` (true/false) (required) and `opts` (optional) [reload options](https://www.npmjs.com/package/reload#table-of-options-for-reload-opts-parameter).
+  - Default: `{enable: true}`
```
kethinov: i see both `enable` and `enabled`

lib/reload.js
kethinov: 👍

lib/reload.js
```diff
@@ -0,0 +1,7 @@
...
+  var reload = require('reload');
```
kethinov: don't require it unless the if statement passes

kethinov:
* Port Stuff
  * Default port is fine 
  * No need for us to specify a new one 
* README Stuff
  * Yes, provide a code example somewhere in the README, and we should open an issue on the generator to optionally (on by default) include reload in dev mode in the layout template
* Module Stuff
  * Works for me
* roosevelt config options
  * Works for me

README.md
alallier: Which do you want?

kethinov: `enabled`

alallier:
So something to note here (may want to track this in an issue, that's why I bring it up).

There is a penitential problem with reload. (One of which was originally brought up during the reload re-write which raises it back up to me, anyways...)

Two things:

* First, is that reload provides no callback when it is done, so it is arbitrarily placed before map routes but it has to happen before the mapping of routes otherwise it will get placed at the end and the route   for `reload/reload.js` will not get created. There is a proposal for reload providing a callback when it is all done (see: alallier/reload#93). If this gets added to reload it should be added here
* Second, as soon as reload is called it creates the WebSocket server signaling to the browser to reload, but considering where in the chain reload is called this could happen before the server is started. I'm thinking reload will need a second method that when called will initialize the WebSocket server, so we can call that inside of our `startServer`, otherwise we can have a race condition if some other big thing happened _after_ `mapRoutes` but _before_ `startServer`

alallier:
Addressed the above [comment](https://github.com/rooseveltframework/roosevelt/pull/193#issuecomment-314465612) in 93718480da6bf43c1aafb92f9be3b83d25387a1e

kethinov:
the merge conflicts are so bad on this by now that it probably makes more sense to just start over with a fresh PR

PR #199, opened 2017-07-17
Fixes handling of complex data structures in parameters (Closes #189)

### LINKED ISSUE #189: Parameter priority - complex data structures
Checking for existence is not enough for parameters with things like objects and arrays. See original comment made on this [here](https://github.com/rooseveltframework/roosevelt/issues/168#issuecomment-312087813)

To re-iterate:

Pick a roosevelt param that is an object we will use `validatorExceptions`. This object accepts two possible keys `requestHeader` and `modelValue`.

If I was to specify the `validatorExceptions` object in the `package.json` with only one key like:

```json
{"requestHeader": "Partial"}
```

Source params does this check:

```javascript
params.validatorExceptions !== undefined ? params.validatorExceptions : pkg.rooseveltConfig.validatorExceptions !== undefined ? pkg.rooseveltConfig.validatorExceptions : defaultConfig.validatorExceptions
```

So `validatorExceptions` is defined in the `package.json` but not **all** keys are defined. But we can never move onto the defaults because it is present in the `package.json` so `modelValue` gets set to null. This can cause problems. Objects and arrays will need more stringent checks.

dishad:
Created a function that handles objects and other regular parameters. 

Closes #189


Autre31415:
@kethinov What do you think of renaming `lib/checkFiles.js` to something more generic like `common.js` and moving the `checkObject` and `checkParams` functions into it?

kethinov:
i'd go with `utils.js` but otherwise sounds good

Autre31415:
In that case @dishad when you do that be sure to replace any instance of `checkFiles` with `utils` wherever it's required across the codebase.

dishad:
Sounds good ill update this PR

Autre31415:
LGTM

PR #224, opened 2017-10-10
Crash the app with a human readable error if roosevelt-closure is too out of date

Autre31415:


lib/jsCompiler.js
```diff
@@ -26,6 +30,25 @@ module.exports = function(app, callback) {
     return;
   }
 
+  if (preprocessor === 'roosevelt-closure') {
```
kethinov: hardcoding is bad

this should check if the api of the dependency module is out of date, not check for the specific module

kethinov: an even better solution would be designing a system that examines the dependency module to ensure that it is broadly compatible with the current version of roosevelt. if it's not compatible that could be for many reasons, such as it being too old, but also because someone cited their js compiler module as something totally weird, like, say, teddy instead of roosevelt-closure

PR #238, opened 2017-10-30
Add priority levels to script audit

### LINKED ISSUE #237: Config auditor will complain about missing npm scripts that we should probably consider optional
e.g. remove the `eslint`, `stylelint`, or `lint` scripts, and the config auditor will complain that they are missing.

Strictly speaking, those (and possibly other things?) should be considered optional.

There is also probably a class of optional *but recommended* things too. I can't think of any right now, but in such cases, I would degrade this error to a warning, and explicitly note that roosevelt recommends that it be present even though it is optional.

Basically two extra layers: a class of optional configs that roosevelt shows a warning for when it's missing, and a class of optional configs that roosevelt will ignore when it's missing without showing an error or a warning.

The whole config auditor needs to be audited to identify configs which belong on these reduced levels of urgency.

Autre31415:
Priority level selection is up for discussion.

Closes #237

### DIFF
--- a/lib/configAuditor.js
+++ b/lib/configAuditor.js
@@ -30,10 +30,10 @@ require('colors');
 
   defaultConfigKeys.forEach(function (defaultParam) {
     if (userConfig[defaultParam] === undefined) {
-      console.log(`⚠️   Missing param ${defaultParam}!`.red.bold)
+      console.error(`⚠️   Missing param ${defaultParam}!`.red.bold)
       errors = true
     } else if (typeof defaultConfig[defaultParam] !== typeof userConfig[defaultParam]) {
-      console.log(`⚠️   Param ${defaultParam} structured incorrectly, should be ${typeof defaultParam}`.red.bold)
+      console.error(`⚠️   Param ${defaultParam} structured incorrectly, should be ${typeof defaultParam}`.red.bold)
       errors = true
     } else if (defaultConfig[defaultParam] !== (undefined || []) && defaultConfig[defaultParam] instanceof Object) {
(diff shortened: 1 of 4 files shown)


lib/configAuditor.js
```diff
@@ -51,8 +51,12 @@ require('colors');
 
   defaultScriptKeys.forEach(function (defaultScript) {
     if (userScripts[defaultScript] === undefined) {
-      console.log(`⚠️   Missing script ${defaultScript}!`.red.bold)
-      errors = true
+      if (defaultScripts[defaultScript].priority === 'error') {
+        console.log(`⚠️   Missing script ${defaultScript}!`.red.bold)
```
kethinov: why not console.error here?

lib/configAuditor.js
```diff
@@ -51,8 +51,12 @@ require('colors');
...
+        errors = true
+      } else if (defaultScripts[defaultScript].priority === 'warning') {
+        console.log(`⚠️   Missing recommended script ${defaultScript}!`.yellow.bold)
```
kethinov: and console.warn here?

lib/defaultScripts.json
```diff
@@ -1,13 +1,46 @@
...
+    "value": "./node_modules/.bin/stylelint \"statics/css/**/*.less\" --syntax less",
+    "priority": "ignore"
+  },
+  "lint": {
+    "value": "npm run stylelint && npm run eslint",
+    "priority": "ignore"
+  },
+  "precommit": {
+    "value": "lint-staged",
+    "priority": "error"
```
kethinov: ignore

Issue #245, opened 2017-11-08
Enhancement: HTTPS CA Chain Loading


jkae:
There will, occasionally, be a case where a user would like to create an HTTPS server through Roosevelt -- but instead of a single/multiple file/s that contain a single certificate each, they have one .pem file containing a full chain of certificates.

As it is now, Roosevelt accepts two formats as part of the `config.ca` option:

1. A file path containing a single certificate
2. An array of file paths, each containing a single certificate

We have the following options:

1. If a string (path to a single file) is given as the ca option, we can read it as normal but run it through an additional parsing library like `split-ca`.
2. If reading a file throws `ENOENT`, we can make two assumptions:
 a. The user has passed us a certificate string.
 b. The user has passed us a bad filename.

I don't know how acceptable option 1 is, but 2 seems viable.

In the case of 2a this can either be captured on a case-by-case basis in the loop through `ca[]`, or assumed that the entire array is certificate strings.

In the case of 2b, if we simply set `httpsOptions.ca` to the given value, it's going to throw when attempting to start the server. Since the current behavior throws an error anyway, I don't think that's too big a deal.

kethinov:
Published as 0.9.3.

PR #246, opened 2017-11-08
HTTPS CA / Accept fully-realized string input

jkae:
#245 - Implementation of discussed option 2

- try/catch around reading CA files
- assume string keys if File Not Found
  - if any fail, apply assumption to entire array
- allow https server to deal with bad CA input

This is just a quick and dirty solution, but I'm willing to go with whatever implementation might be desired.

### DIFF
--- a/roosevelt.js
+++ b/roosevelt.js
@@ -78,14 +78,22 @@ module.exports = function (params) {
         httpsOptions.passphrase = passphrase
       }
       if (ca) {
-        // String or array
-        if (typeof ca === 'string') {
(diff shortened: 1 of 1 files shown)


kethinov:
You should close this and reopen it with a commit bound to the email address that is actually associated with your GitHub account.

jkae:
I thought it was but it turns out I was wrong.

kethinov:
Sorry when I said reopen, I meant a fresh PR without the tainted history.

alallier:
You could always re-write the authorship and force push the branch

Issue #249, opened 2017-11-08
[minor] HTTPS-Only still attempts to add listener to httpServer


jkae:
**roosevelt.js:106..115**
```
// assign individual keys to connections when opened
httpServer.on('connection', function (conn) {
  let key = conn.remoteAddress + ':' + conn.remotePort
  connections[key] = conn

  // once the connection closes, remove
  conn.on('close', function () {
    delete connections[key]
  })
})
```

When running Roosevelt in `httpsOnly: true`, `httpServer` will not be assigned a value. This call to `httpServer.on` will fail & throw an error.

I also find it interesting that this listener is only active on httpServer connections & ignores httpsServer, though that is likely intentional.

kethinov:
That's definitely a bug.

PR #250, opened 2017-11-10
Fix / http connection mapping

jkae:
Resolves issue #249 

- Moves anonymous function to declared `mapConnections`
- Sets up listener on `httpServer` inside `!httpsOnly` block
- Adds listener to `httpsServer` inside `https` block


jkae:
I think Visual Studio Code has some problems when moving lines around.

PR #258, opened 2017-11-21
Parameter: CA Is File

### LINKED ISSUE #257: bug: ENAMETOOLONG, MacOS
### Problem
1. MacOS -- and possibly GNU/Linux -- has a maximum filename length that cases `fs` to throw `ENAMETOOLONG`. Thus, on non-Windows operating systems, we cannot try-catch a string CA without also including that error code.
2. It would be entirely possible that someone typos a filename, uses a deleted file, or has the wrong path to a file, for their CA. In that case, we would want to fail immediately so that the developer would be notified of the error.

### Solution
- Replaces try-catch `fs` read with an if-else determined by a new parameter, `cafile`.
  - Defaults to `true`, suggesting that `ca` is a file path.
  - If `false`, `ca` is assumed to be a certificate string already, and used as-is.

jkae:
Implementation of solution.

Closes #257

### DIFF
--- a/README.md
+++ b/README.md
@@ -290,8 +290,10 @@ HTTPS parameters
   - Default: `null`
 - `passphrase`: Supply the HTTPS server with the password for the certificate being used, if necessary.
   - Default: `null`
-- `ca`: Certificate authority to match client certificates against, as a file path or array of file paths.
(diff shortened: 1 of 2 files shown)


kethinov:
released in 0.10.1.

PR #270, opened 2017-11-29
Overhaul logging and config auditor

Autre31415:
With the addition of the `suppressLogs` feature giving users the ability to hide roosevelt logs we are required to check the state of it before each console.log. It was ugly doing this inline all across the module, so now we've got an internal logging system!

`logger` is a simple class that gives us custom functions for console.log, console.warn, and console.error. `logger.log` checks the state of `suppressLogs` is on before logging, `logger.warn` appends a ⚠️ to the message and runs console.warn, and `logger.error` appends ❌ and runs console.error.

In addition I've also made some big changes to the `defaultParams` object to include metadata about each param (Including the name, if it's optional, and parent if any). This change is mainly to give more power to the config auditor, and make it much easier to adapt big changes to the API down the road.

I've also attempted to fix all of the odd emoji spacing errors in the console output with the goal of course being consistent spacing across each message.


lib/logger.js
```diff
@@ -0,0 +1,19 @@
...
+  var suppressLogs = app.get('params').suppressLogs.rooseveltLogs
+
+  this.log = function () {
+    if (!suppressLogs) {
+      console.log(...arguments)
+    }
+  }
+
+  this.warn = function () {
+    console.log('⚠️ ', ...arguments)
```
kethinov: this should be `console.warn`

lib/logger.js
```diff
@@ -0,0 +1,19 @@
...
+  }
+
+  this.error = function () {
+    console.log('❌ ', ...arguments)
```
kethinov: this should be `console.error`

PR #296, opened 2017-12-18
WIP: Process to kill detached validator after set time

JoshL18:
The goal of this change is in the case that a developer kills their roosevelt app but forgets to run the command to kill the validator (`npm run kill-validator`), a background process will kill it for them.

With this change, upon starting a detached validator, a background process is spawned which:

- Begins a timer
- Once time expires, background process kills the validator, then exits itself.

The time is reset in dev mode only by:
- start/restart of any roosevelt server
- Roosevelt apps set an interval to refresh timer process so validator will not get killed accidentally


kethinov:
Closing this as it appears the work has been abandoned and by this point the merge conflicts are so thick you'd need to open a new one anyway.

PR #306, opened 2018-01-02
WIP: Overhaul Roosevelt params

### LINKED ISSUE #194: Merge possible config items together
Possible enhancement, merge possible config items together. Right now we split up enabling/disabling of a feature, with it options. Example below:

```JSON
"enableValidator": true,
"htmlValidator": {
  "port": "8888",
  "format": "text"
}
```

_Same with browserify bundling and many others._

I think it might be cleaner to put the enabling/disabling in the same object as the options config.

So the above example would be:

```JSON
"htmlValidator": {
  "enable": true,
  "options": {
    "port": "8888",
    "format": "text"
  }
}
```

So in this new case you would now pass `htmlValidator.options` in place of `htmlValidtor`, which results in the same object being passed as before.

This cuts down on the number of config options and makes it very clear how to enable and configure options. 

What are people's thoughts on this?

Note: This was originally thought of in #193

### LINKED ISSUE #294: Refactor npm scripts for consistency, stylistic conformance to common patterns, and add shorthands
We currently have:

- `start`
- `prod`
- `dev`
- `kill-validator`
- `cleanup`
- `audit`
- `eslint`
- `stylelint`
- `lint`
- `precommit`
- `test`

It should change to:

- `start`
- `production`
  - Shorthand: `prod`
  - Shorthand: `p`
- `development`
  - Shorthand: `dev`
  - Shorthand: `d`
- `kill-validator`
  - Shorthand: `kv`
- `clean`
  - Shorthand: `c`
- `audit`
  - Shorthand: `a`
- `standard` (replace `eslint`)
  - Just runs `standard`. A bit silly to have an npm script more verbose than the native command line tool available, but the utility is to surface that the tool is available when the user runs `npm run` to list the scripts available.
- `stylelint`
- `lint`
  - Shorthand: `l`
  - Also change this to run standard instead of eslint.
- `precommit`
  - Shorthand: `pc`
- `test`
  - Shorthand: `t` (this is natively implemented by npm so no need to specify it)

This change will only be merged when we're ready to release 0.11.0.

Autre31415:
This is the big one that finally addresses the necessary quality of life updates to the user param structure. Sweeping changes are made. (Closes #194)

- `enableValidator`, and `validatorExceptions` have been made children of `htmlValidator`.
  - Renamed to `enable`, and `exceptions`.
- New param `css`. The following have been made children of it:
  - `cssPath`, `cssCompiler`, `cssCompilerWhitelist`, `cssCompiledOutput`, and `versionedCssFile` renamed to `sourceDir`, `compiler`, `whitelist`, `output`, and `versionFile`.
- New param `js`. The following have been made children of it:
  - `jsPath`, `jsCompiler`, `jsCompilerWhiteList`, `jsCompilerBlacklist`, and `jsCompiledOutput` renamed to `sourceDir`, `compiler`, `whitelist`, `blacklist`, and `output`.
  - New param `bundler`. The following have been made children of it:
    - `browserifyBundles`, `bundledJsPath`, and `exposeBundles` renamed to `bundles`, `output`, and `expose`.
- `symlinksToStatics` renamed to `staticsSymlinksToPublic`.
- `versionedStatics` renamed to `versionedPublic`.
- `nodeEnv` param now properly overrides `NODE_ENV` and express `env` variable.
- Brought `defaults/config.json` back to a simple structure. This makes it easier to use, and also allows it to be used as a reference to the current API.
- Add more default npm scripts. (Closes #294)
- Added inspection of scripts that utilize built in files to the config auditor.
- Rename `cleanup` script to `clean`.

### DIFF
--- a/lib/defaults/scripts.json
+++ b/lib/defaults/scripts.json
@@ -36,7 +36,7 @@
     "value": "npm run kill-validator",
     "priority": "ignore"
   },
-  "cleanup": {
+  "clean": {
     "value": "node ./node_modules/roosevelt/lib/scripts/appCleanup.js",
     "priority": "error",
     "inspect": true
--- a/lib/defaults/scripts.json
+++ b/lib/defaults/scripts.json
@@ -1,46 +1,85 @@
 {
   "start": {
-    "value": "npm run prod",
+    "value": "npm run production",
     "priority": "error"
   },
-  "kill-validator": {
-    "value": "node ./node_modules/roosevelt/lib/killValidator.js",
+  "production": {
+    "value": "nodemon app.js --production-mode",
     "priority": "error"
   },
   "prod": {
-    "value": "nodemon app.js -prod",
+    "value": "npm run production",
+    "priority": "ignore"
+  },
+  "p": {
+    "value": "npm run production",
+    "priority": "ignore"
+  },
+  "development": {
+    "value": "nodemon app.js --development-mode",
     "priority": "error"
   },
   "dev": {
-    "value": "nodemon app.js -dev",
-    "priority": "error"
+    "value": "npm run development",
+    "priority": "ignore"
+  },
+  "d": {
+    "value": "npm run development",
+    "priority": "ignore"
+  },
+  "kill-validator": {
+    "value": "node ./node_modules/roosevelt/lib/scripts/killValidator.js",
+    "priority": "error",
+    "inspect": true
+  },
+  "kv": {
+    "value": "npm run kill-validator",
+    "priority": "ignore"
   },
   "cleanup": {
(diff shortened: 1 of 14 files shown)


lib/defaults/scripts.json
```diff
@@ -1,46 +1,85 @@
...
+    "inspect": true
+  },
+  "kv": {
+    "value": "npm run kill-validator",
+    "priority": "ignore"
   },
   "cleanup": {
-    "value": "node ./node_modules/roosevelt/lib/appCleanup.js",
-    "priority": "error"
+    "value": "node ./node_modules/roosevelt/lib/scripts/appCleanup.js",
```
kethinov: we need to deprecate cleanup and replace it with clean

PR #329, opened 2018-01-15
Command line test

siuster95:


test/commandLine/commandLineTest.js
```diff
@@ -0,0 +1,662 @@
...
+    const testApp = fork(path.join(appDir, 'app.js'), ['-dev'], {'stdio': ['pipe', 'pipe', 'pipe', 'ipc']})
+
+    testApp.on('message', params => {
+      assert.equal(params.nodeEnv, 'development')
+      done()
+    })
+  })
+
+  it('should change the app to development mode ("-development-mode")', function (done) {
+    const testApp = fork(path.join(appDir, 'app.js'), ['-development-mode'], {'stdio': ['pipe', 'pipe', 'pipe', 'ipc']})
```
Autre31415: this should be `--development-mode`

test/commandLine/commandLineTest.js
```diff
@@ -0,0 +1,662 @@
...
+    rimraf(appDir, (err) => {
+      if (err) {
+        throw err
+      } else {
+        done()
+      }
+    })
+  })
+
+  /* check individual flags */
```
Autre31415: Add a category for soon to be deprecated flags and move them all up there together. That way when we deprecate them we can dump their tests too without having to dig. Also avoid using soon to be deprecated flags in the combination tests for this same reason.

test/commandLine/commandLineTest.js
```diff
@@ -0,0 +1,662 @@
...
+  it('should change the app to attach the html Validator ("-a")', function (done) {
+    const testApp = fork(path.join(appDir, 'app.js'), ['-a'], {'stdio': ['pipe', 'pipe', 'pipe', 'ipc']})
+
+    testApp.on('message', params => {
+      assert.equal(params.htmlValidator.separateProcess, false)
+      done()
+    })
+  })
+
+  /* check combination of flags */
```
Autre31415: Since the first group of tests prove the functionality of each flag along with each alias we don't need to be quite so verbose in combination tests. Testing combinations with a single version of each flag will suffice.

Example: If we test `--development-mode --enable-validator --attach-validator` we don't need to also test `--dev --html-validator -a`.

Instead let's cut it down to just unique combinations.

This applies to mutually exclusive flags as well.

test/commandLine/commandLineTest.js
```diff
@@ -0,0 +1,662 @@
...
+
+    testApp.on('message', params => {
+      assert.equal(params.nodeEnv, 'production')
+      assert.equal(params.htmlValidator.enable, false)
+      assert.equal(params.htmlValidator.separateProcess, true)
+      done()
+    })
+  })
+
+  /* check one letter combinations */
```
Autre31415: Add some more combinations of single letter params here like `-had` or `-bap`.

test/commandLine/commandLineTest.js
```diff
@@ -0,0 +1,662 @@
...
+
+    testApp.on('message', params => {
+      assert.equal(params.nodeEnv, 'production')
+      assert.equal(params.htmlValidator.enable, false)
+      assert.equal(params.htmlValidator.separateProcess, true)
+      done()
+    })
+  })
+
+  /* check mutually exclusive flags  */
```
Autre31415: Add a test to this section that tests the limits of mutually exclusive flags like `-dpabhr`

test/commandLine/commandLineTest.js
```diff
@@ -0,0 +1,664 @@
...
+    const app = require('../../../roosevelt')({
+      appDir: appDir,
+      generateFolderStructure: false,
+    })
+    if (process.send) {
+      process.send(app.expressApp.get('params'))
+    }
+
+    process.exit()
+    `
```
kethinov: i don't think we should write this code inline like this. it's pretty hard to read and maintain. put it in a separate file somewhere appropriate in the `test` directory somewhere, then open the file here and dump its contents into the `writeFileSync`

test/commandLine/commandLineTest.js
```diff
@@ -0,0 +1,664 @@
...
+
+    fs.writeFileSync(path.join(appDir, 'app.js'), dataIn)
+  })
+
+  after(function (done) {
+    rimraf(appDir, (err) => {
```
kethinov: we should abstract this into a utility call so that there is one call to `rimraf` across all tests. @Autre31415 down the road we might want to make that utility function *absolutely filled* with defensive logic to ensure it doesn't delete anything that it shouldn't be deleting

test/commandLine/commandLineTest.js
```diff
@@ -0,0 +1,609 @@
...
+    testApp.on('message', params => {
+      assert.equal(params.nodeEnv, 'development')
+      assert.equal(params.htmlValidator.enable, true)
+      assert.equal(params.htmlValidator.separateProcess, false)
+      done()
+    })
+  })
+
+  it('should change the app to put it into dev mode, enable validator, and attach validator ("-had")', function (done) {
+    const testApp = fork(path.join(appDir, 'app.js'), ['-dha'], {'stdio': ['pipe', 'pipe', 'pipe', 'ipc']})
```
Autre31415: The argument being passed here doesn't match the description of the test

test/commandLine/commandLineTest.js
```diff
@@ -0,0 +1,609 @@
...
+
+    testApp.on('message', params => {
+      assert.equal(params.nodeEnv, 'production')
+      assert.equal(params.htmlValidator.enable, false)
+      assert.equal(params.htmlValidator.separateProcess, true)
+      done()
+    })
+  })
+
+  it('should change the app to put it into production mode, disable validator, and detach validator ("-rbp")', function (done) {
```
Autre31415: This argument also doesn't match the test description

test/commandLine/commandLineTest.js
```diff
@@ -0,0 +1,609 @@
...
+
+    testApp.on('message', params => {
+      assert.equal(params.nodeEnv, 'production')
+      assert.equal(params.htmlValidator.enable, false)
+      assert.equal(params.htmlValidator.separateProcess, false)
+      done()
+    })
+  })
+
+  /* check mutually exclusive flags  */
```
Autre31415: Similar to the combinations tests we don't need to check `--prod --dev` if we already checked `--production-mode --development-mode`

Condense these mutually exclusive tests down to unique combinations of flags.

test/commandLine/app.txt
```diff
@@ -0,0 +1,10 @@
+const path = require('path')
```
Autre31415: rename this to `app.js`, move it to the `test/lib` directory, and ensure that the code indentation is 2 spaces and not 4.

test/commandLine/commandLineTest.js
```diff
@@ -0,0 +1,519 @@
...
+    rimraf(appDir, (err) => {
+      if (err) {
+        throw err
+      } else {
+        done()
+      }
+    })
+  })
+
+  /* checking soon to be deprecated flags */
```
Autre31415: These commented category names could be individual describe blocks, that way we can see them in the mocha console output.

PR #334, opened 2018-01-17
first test for js section

siuster95:


test/unit/jsTest.js
```diff
@@ -0,0 +1,234 @@
...
+    path.join(appDir, 'statics', '.build', 'js', 'c.js')
+  ]
+  let staticJSFiles = [
+    test1,
+    test2,
+    test3
+  ]
+
+  // function to compile the static folder and files
+  let generateStaticFolder = () => {
```
Autre31415: Switch this to a function declaration (i.e. `function generateStaticsFolder () {`

test/unit/jsTest.js
```diff
@@ -0,0 +1,234 @@
...
+    })
+  })
+
+  it('should compile all static js files using roosevelt-uglify', function (done) {
+    generateStaticFolder()
+    // create and init the app
+    generateTestApp({
+      appDir: appDir,
+      generateFolderStructure: true,
+      suppressLogs: {
```
Autre31415: Remove this suppressLogs param. When you fork processes in this way the logs aren't printed to the console unless you force them to be.

test/unit/jsTest.js
```diff
@@ -0,0 +1,234 @@
...
+        rooseveltLogs: true,
+        rooseveltWarnings: true
+      },
+      js: {
+        compiler: {
+          nodeModule: 'roosevelt-uglify',
+          showWarnings: false,
+          params: {}
+        },
+        whitelist: ['a.js', 'b.js', 'c.js']
```
Autre31415: There's no need to specify a whitelist outside of the whitelist test. By default all files in `js` are compiled. The whitelist is used when you want to only compile specific files.

test/unit/jsTest.js
```diff
@@ -0,0 +1,234 @@
...
+        let test = pathOfWhiteListedFiles.includes(file.path)
+        assert.equal(test, true)
+      })
+      testApp.kill()
+      done()
+    })
+  })
+
+  it.skip('should minify all files except for those that are blacklisted', function (done) {
+    // change the data that would be put into the static files
```
Autre31415: Why is this data being changed in this test?

test/unit/jsTest.js
```diff
@@ -0,0 +1,234 @@
...
+      assert.equal(test2, false)
+      assert.equal(test3, true)
+      testApp.kill()
+      done()
+    })
+  })
+
+  it('should make the output compiled folder with the new name and put all the compiled JS in it', function (done) {
+    // make everything back to normal
+    staticJSFiles[0] = 'var a = 7'
```
Autre31415: There should be no reason to change these variables inside of tests. They should be specified in the original array and stay the same for the entire suite.

test/unit/jsTest.js
siuster95: Oh yea for the blacklisted tests, you said that if the js files were too simple, their could be a chance that the minified version and the regular version would look the same, so for that test, I made each file a little bit more compicated

Autre31415: No reason to not use the more complex js test files for every test

Autre31415:
Add another test that utilizes the `:` delimiter feature of whitelist to write another file name for the output file. Even better if it's in a directory too. i.e., whitelist: `['a.js:test/hello.js']` should compile `a.js` and output it to `test/hello.js`.

package.json
```diff
@@ -39,9 +39,14 @@
     "eslint": "~4.15.0",
     "eslint-plugin-no-only-tests": "~2.0.0",
     "husky": "~0.14.2",
+    "i": "^0.3.6",
```
Autre31415: What's going on with these added dependencies? `i`, `json-cycle`, and `npm`

PR #352, opened 2018-01-31
WIP: Routes Unit Tests

siuster95:


package.json
```diff
@@ -32,7 +32,8 @@
     "serve-favicon": "~2.4.0",
     "tamper": "~1.0.0",
     "toobusy-js": "~0.5.0",
-    "vnu-jar": "~17.11.0"
+    "vnu-jar": "~17.11.0",
+    "teddy": "~0.4.23"
```
Autre31415: Should be a dev dependency

test/unit/routesTest.js
```diff
@@ -0,0 +1,136 @@
...
+    }, 'startServer')
+
+    // create a fork of the app.js file and run it as a child process
+    const testApp = fork(path.join(appDir, 'app.js'), {'stdio': ['pipe', 'pipe', 'pipe', 'ipc']})
+
+    // on the message sent by the app completing the start method, send a http request for the teddy page to the server and see if it sends back a response
+    testApp.on('message', () => {
+      request('http://localhost:43711')
+      .get('/teddyTest')
+      .expect(200, (err) => {
```
Autre31415: You should also do some checking of the response to ensure that the page rendered properly. Check that the page contains the values you assigned to the teddy variables to prove that.

test/unit/routesTest.js
```diff
@@ -0,0 +1,136 @@
...
+    }, 'startServer')
+
+    // create a fork of the app.js file and run it as a child process
+    const testApp = fork(path.join(appDir, 'app.js'), {'stdio': ['pipe', 'pipe', 'pipe', 'ipc']})
+
+     // on the message sent by the app completing the start method, send a http request for the plain html page to the server and see if it sends back a response
+    testApp.on('message', () => {
+      request('http://localhost:43711')
+      .get('/HTMLTest')
+      .expect(200, (err) => {
```
Autre31415: Also do some response sniffing here to ensure the page loaded properly. In this case check that your statically defined stuff like 'Title' and 'Sentence' is in the response body.

test/util/mvc/controllers/plainHTMLController.js
```diff
@@ -0,0 +1,7 @@
+module.exports = (app) => {
+  // route to request the plain HTML page and return the HTML page
+  app.route('/HTMLTest').get((req, res) => {
+    // render the teddy template and pass it the model
+    res.send('plainHTMLTest')
```
Autre31415: For a `res.send` you need to read the HTML file itself. The best way to do this would be caching a variable:

(require `fs` and `path` outside of the module.exports)

`const plainHTMLPage = fs.readFileSync(path.join(app.get('viewsPath'), plainHTMLTest.html))`

PR #353, opened 2018-02-01
WIP: generateTestApp and cleanupTestApp refactor

siuster95:


test/util/generateTestApp.js
```diff
@@ -5,14 +5,17 @@ const fse = require('fs-extra')
 const path = require('path')
 const util = require('util')
 
-module.exports = function (params, method) {
+module.exports = function (params, rooseveltPath, method) {
```
Autre31415: Rather than making three arguments `rooseveltPath` and `method` should be members of an options object. That way in the future it will be easier to add more configuration possibilities. When you do this also rename `params` to `rooseveltParams` for clarity.

test/util/generateTestApp.js
```diff
@@ -5,14 +5,17 @@ const fse = require('fs-extra')
...
   const appDir = params.appDir
-  let appJSContents = `const app = require('../../../roosevelt')(${util.inspect(params, {depth: 5})})\n\n`
+  let appJSContents = `const app = require(${util.inspect(rooseveltPath)})(${util.inspect(params, {depth: 5})})\n\n`
```
Autre31415: You shouldn't need util for a string

PR #354, opened 2018-02-02
Expose more dependencies as express variables (Closes #351)

### LINKED ISSUE #351: Expose references to all dependency modules as Express variables
Some are exposed now, but not all, e.g. cookie-parser.

Autre31415:


kethinov:
On second thought these don't really need to be exposed in order for the dependency modules to be accessed and useful outside of their default contexts.

PR #361, opened 2018-02-09
Parameter functions unit test (quest to 100% with mapRoutes, coveralls: 100%)

siuster95:


test/unit/paramFunctionTest.js
```diff
@@ -0,0 +1,140 @@
...
+      }
+    })
+  })
+
+  it('should execute what is in onReqStart', function (done) {
+    // generate the app
+    generateTestApp({
+      appDir: appDir,
+      generateFolderStructure: true,
+      onReqStart: true,
```
Autre31415: I understand why you're using `true` for these functions and then overriding them in the app.js generator, but I'm concerned that we're piling on technical debt at this point. It would be ideal to be able to custom tailor these functions in tests, and if we ever need to then we'd have to go back to all of these old tests and change them later on. We should find a way to pass functions here that get written to the file directly without overrides.

PR #362, opened 2018-02-12
WIP: refactor app kill signal and the ending of the tests

siuster95:


test/unit/routesTest.js
```diff
@@ -310,6 +331,7 @@ describe('Roosevelt routes Section Test', function () {
     testApp.on('message', () => {
       // indicator to tell whether or not we had a 503 come
       let bool503 = false
+      let killSignalSent = false
```
Autre31415: Could you explain what's going on here?

test/unit/routesTest.js
```diff
@@ -308,28 +323,39 @@ describe('Roosevelt routes Section Test', function () {
...
-            // check to make sure that all specific pharses are there
-            assert.equal(test1, true)
-            assert.equal(test2, true)
-            assert.equal(test3, true)
-            testApp.kill()
-            done()
-          }
-        })
-      }
+      async.series([
```
kethinov: i don't think we need a new dep here. there are native promises in node now for handling async workflows

test/unit/routesTest.js
```diff
@@ -308,28 +323,39 @@ describe('Roosevelt routes Section Test', function () {
...
+        function (cb) { request('http://localhost:43711').get('/HTMLTest').expect(200, (err, res) => { if (err) { cb(res, null) } else { cb(null, res) } }) },
```
kethinov: also i think there is probably a more intelligent way to structure this since each request is the same, can't we just fire off a loop that generates a bunch of these rather than copy/paste?

test/unit/routesTest.js
```diff
@@ -348,30 +374,40 @@ describe('Roosevelt routes Section Test', function () {
...
+        function (cb) { request('http://localhost:43711').get('/HTMLTest').expect(200, (err, res) => { if (err) { cb(res, null) } else { cb(null, res) } }) },
+        function (cb) { request('http://localhost:43711').get('/HTMLTest').expect(200, (err, res) => { if (err) { cb(res, null) } else { cb(null, res) } }) }
+      ], (err, result) => {
+        // if any one of the test brings an error. (one of them should). Check to see if the err response text has the custom 503 page we put in there
+        if (err) {
+          // sample the response text
+          const test1 = err.res.text.includes('503 Service Unavailable')
+          const test2 = err.res.text.includes('The requested URL /HTMLTest is temporarily unavailable at this time')
+          const test3 = err.res.text.includes('localhost:43711')
```
kethinov: probably don't want to hardcode the port in here in case the default port ever changes

PR #364, opened 2018-02-13
public parameters Unit Tests 

siuster95:


test/unit/publicFolderTest.js
```diff
@@ -0,0 +1,157 @@
...
+    cleanupTestApp(appDir, (err) => {
+      if (err) {
+        throw err
+      } else {
+        done()
+      }
+    })
+  })
+
+  it('should be able to send a web page back with no problem with favicon set to something', function (done) {
```
Autre31415: In this test could you look for a way to prove that the favicon is loaded with the page?

test/unit/publicFolderTest.js
```diff
@@ -0,0 +1,157 @@
...
+        testApp.kill('SIGINT')
+      }
+    })
+
+    testApp.on('exit', () => {
+      done()
+    })
+  })
+
+  it('should allow the user to name the symlink files in the public folder', function (done) {
```
Autre31415: I believe we already have a test for this in the folder tests

test/unit/publicFolderTest.js
Autre31415: Also, once you have that figured out add 2 more tests for maximum coverage:
- Have favicon set  to null and test that no favicon is loaded
- Have the favicon set to the wrong path where there is no image and test that no favicon is loaded

test/unit/publicFolderTest.js
```diff
@@ -0,0 +1,226 @@
...
+      .get('/HTMLTest')
+      .expect(200, (err, res) => {
+        if (err) {
+          assert.fail(err)
+          testApp.kill('SIGINT')
+        }
+        // if we can get the page, send a request to get the favicon
+        request('http://localhost:43711')
+        .get('/favicon.ico')
+        .expect(200, (err, res) => {
```
Autre31415: Since you're expecting this to fail, why not expect a 404?

PR #365, opened 2018-02-13
https config unit tests

jkae:
- Testing for various states of https config
  - Most invalid states cause expected issues in `fs`, `fse` and `https`; I don't think it's fully in-scope to test unless there is specific error handling in the future
  - Contains test config, test .key, .crt, and .p12 files
  - Utilizes `sinon` and `proxyquire` for test spying to ensure correct calls are being made
- Small edit to various `app.get('params').https.keyPath` calls -- now stores as variable so its properties are more-easily accessed


test/unit/httpsOptions.js
```diff
@@ -0,0 +1,154 @@
...
+    app({appDir: appDir, ...config})
+    assert(stubHttpsServer.args[0][0].ca[0].equals(ca1), 'https.Server CA (1) did not match supplied CA')
+    assert(stubHttpsServer.args[0][0].ca[1].equals(ca2), 'https.Server CA (2) did not match supplied CA')
+
+    config.https.ca = 'test/util/ca.crt'
+  })
+
+  it('should pass certificate authority directly to httpsServer if cafile is false', function () {
+    config.https.cafile = false
+    config.https.ca = '-----BEGIN CERTIFICATE-----\n' + // I don't think template strings work here
```
Autre31415: Since this matches the contents of `ca.crt` couldn't you use `fs.readFileSync` to get the string?

test/util/testHttpsConfig.json
```diff
@@ -0,0 +1,86 @@
+{
```
Autre31415: What's going on in this config file?

jkae: I just set everything to a bogus value because I didn't need anything to work. Admittedly it's mostly a waste of space, so I'll just delete everything that I'm not explicitly using.

test/unit/httpsOptions.js
jkae: That's true. It also doesn't even matter what the contents of `config.https.ca` are, here. I could do that or I could just make it "test" or something.

test/util/testHttpsConfig.json
Autre31415: We're pretty rigorous about setting defaults. You could just pass the needed https params to the constructor, that should make it a lot simpler.

PR #387, opened 2018-03-01
 htmlValidator and killValidator 100%

siuster95:
implements #404


lib/scripts/killValidator.js
```diff
@@ -88,66 +92,86 @@ getValidator()
...
+          scanPort(currentPortinTest, endingPort)
+        })
+
+        socket.on('end', () => {
+          socket.destroy()
+          currentPortinTest++
+          scanPort(currentPortinTest, endingPort)
+        })
+      } else {
+
```
Autre31415: Why is there an empty else condition here?

lib/scripts/killValidator.js
```diff
@@ -88,66 +92,86 @@ getValidator()
...
-
-      socket.connect(currentPort, host, function () {
-        let options = {
-          url: 'http://localhost:',
-          port: currentPort,
-          method: 'GET',
-          headers: {
-            'User-Agent': 'request'
+/*
+MY PROPOSAL: From what I gather, my hypothesis on why the original way was not working was because since it was blasting through trying to connect to 65535 ports at once,
```
Autre31415: This proposal comment needs to be removed before this can be merged

roosevelt.js
```diff
@@ -245,10 +245,36 @@ module.exports = function (params) {
       })
     } else {
       if (!app.get('params').https.httpsOnly) {
-        servers.push(httpServer.listen(app.get('port'), (params.localhostOnly && appEnv !== 'development' ? 'localhost' : null), startupCallback(' HTTP', app.get('port'))))
+        servers.push(httpServer.listen(app.get('port'), (params.localhostOnly && appEnv !== 'development' ? 'localhost' : null), startupCallback(' HTTP', app.get('port'))).on('error', (err) => {
+          if (err) {
```
Autre31415: Store these in a single function and call it twice, rather than repeating code.

PR #394, opened 2018-03-07
ConfigAuditor Quest to 100% (Coveralls 100%)

siuster95:


Autre31415:
Ideally you should be able to test the configAudit script without starting an app. While it is true the script does indeed run in this particular scenario, it does so only incidentally. Roosevelt will run the configAudit script when it thinks you're starting an app for the first time and it determines this by checking for the existence of the public folder which wouldn't exist yet at this point if the app is being started fresh. Since this is designed behavior it should be tested in isolation as well in addition to other tests.

The ideal test suite would look something like this:
1. Test configAudit by itself by requiring it directly.
    - Also try to find a way to suppress its console output from spamming the console.
2. Test configAudit by initializing it as a child process fork. (See if you can manipulate the cwd to be correct in this scenario. I believe the child_process exposes a param for this)
3. Test that configAudit is executed when a testApp is started that doesn't contain a `public` folder (Basically the same as the test already here)
4. Test that configAudit doesn't execute when a testApp is started that does contain the `public` folder.

test/unit/configAuditorTest.js
```diff
@@ -0,0 +1,608 @@
...
+    let testApp = fork(path.join(appDir, '../', '../', '../', '/lib', '/scripts', '/configAuditor.js'), [], {'cwd': appDir, 'stdio': ['pipe', 'pipe', 'pipe', 'ipc']})
+
+    testApp.stdout.on('data', (data) => {
+      if (data.includes('Starting roosevelt user configuration audit...')) {
+        startingConfigAuditBool = true
+      }
+      if (data.includes('Configuration audit completed with no errors found.')) {
+        noErrorsBool = true
+      }
+      console.log(`stdout: ${data}`)
```
Autre31415: console.log here

Autre31415:
:tada: :tada: 5.4%!

PR #405, opened 2018-03-26
Added minimist for command line argument parsing

alallier:
I the parsed command line arguments in the return API and in the Express application. Not sure where you want it, I'm happy to change it given your review.

I probably need to add a test for this?


kethinov:
After discussion with @alallier we decided against this addition for now. Might reconsider in the future to refactor [stuff like this](https://github.com/rooseveltframework/roosevelt/blob/92ed0187a552f02661730c6b6cff931142477e98/lib/sourceFlags.js#L32) and then expose to the dependency to the user in the way this PR does, but for now there is not a sufficient enough need.

PR #409, opened 2018-03-28
Proposal for compressing the test cases

siuster95:


kethinov:
need to fix the merge conflict

package.json
```diff
@@ -48,7 +48,8 @@
     "sinon": "~5.0.0",
     "standard": "~11.0.0",
     "supertest": "~3.0.0",
-    "teddy": "~0.4.23"
+    "teddy": "~0.4.23",
+    "execa": "~0.10.0"
```
kethinov: this should be removed if it is not being used

also need to fix the merge conflict

PR #425, opened 2018-04-17
 roosevelt 100% part 2 

siuster95:


siuster95:
Removed the EPERM and EADDRNOTAVAIL from the server.listen error catch. I am not aware of a way for  them to be triggered in roosevelt, thus I have cut them out

roosevelt.js
```diff
@@ -234,18 +234,12 @@ module.exports = function (params) {
 
     function serverPush (server, serverPort, serverFormat) {
       servers.push(server.listen(serverPort, (params.localhostOnly && appEnv !== 'development' ? 'localhost' : null), startupCallback(` ${serverFormat}`, serverPort)).on('error', (err) => {
-        if (err) {
-          if (err.message.includes('EADDRINUSE')) {
-            logger.error(`Another process is using port ${serverPort}. Either kill that process or change this app's port number.`.red)
-          } else if (err.message.includes('EPERM')) {
```
kethinov: you need to restore this error message

the trouble is we were sniffing for the wrong error code. this should be scanning for EACCES to print this error

your other unit test below that scans for EACCES because it wasn't an error code we were handling can be switched to EPERM instead

kethinov:
merge conflict

PR #432, opened 2018-04-20
killValidator 100% part 2 

siuster95:


siuster95:
Got rid of the try catch blocks like I did for the html Validator

PR #472, opened 2018-06-06
Added Environment Variable for Detaching Validator

### LINKED ISSUE #221: Add environment variable for detaching validator
Useful for setting an environment variable permanently in say a `.bashrc` for example, allowing any Roosevelt to have a detached by default HTML validator. Might be cool to also allow on on by default validator with either the same environment variable or a separate one.

JoshL18:
(closes #221)

This change will allow for the use of an environment variable for defaulting the HTML Validator to a detached or attached state. This change:

- [x] Added existence of "IS_DETACHED" environment variable with true or false value.
- [x] Mentioned ENV var in README section related to validator
- Allows for faster testing and general usage of apps. (i.e. Having the environment variable set to true allows your validator to run in a detached state with only typing `npm run dev` and no additional flags everytime across all shells)

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,10 +177,11 @@ npm run dev -- -attach-validator
 ```
 - `-a` can also be used.
 
-Roosevelt makes use of an ENV variable to set the validator to an attached or detached state by default. 
-- Set an environment variable `ROOSEVELT_IS_DETACHED` to true for a detached by default validator.
-- Set an environment variable `ROOSEVELT_IS_DETACHED` to false for an attached by default validator.
+Roosevelt makes use of an ENV variable to set the validator to an attached or detached state by default.
 
+(Note that when the ENV variable is set, it overrides the settings in package.json and params.) 
+- Set an environment variable `ROOSEVELT_VALIDATOR` to detached for a detached by default validator.
+- Set an environment variable `ROOSEVELT_VALIDATOR` to attached for an attached by default validator.
 
 Configure how many CPUs your app will run on:
 ```
--- a/README.md
+++ b/README.md
@@ -198,6 +198,33 @@ npm run audit
 
 See also the [the full list of default scripts](https://github.com/rooseveltframework/roosevelt/blob/master/lib/defaults/scripts.json).
 
+Setting Up Environment Variable
+===
+
+Roosevelt's HTML Validator can be set up to be ran in the detached or attached state by default through use of the environment variable, `IS_DETACHED`.
+There are a few different ways that you can go about setting up this environment variable, depending on what OS your machine is running:
+
+Windows
+---
+1. Open up your Control Panel and navigate to System and Security, and then System.
+2. Click on 'Advanced System Settings' in the side bar and then the 'Environment Variables' button.
+3. Press 'New'. In the 'Variable name' box, type `IS_DETACHED`.
+4. In the 'Variable value' box, type true if you want your validator detached by default, or false for attached by default.
+5. Click 'Ok' until you are back at the System Properties window, then press 'Apply' and 'Ok'
+6. Your environment variable for detaching/attaching the HTML validator is now set up and ready to be used.
+* If you would like to change the value associated with `IS_DETACHED`, in the command prompt, type `set IS_DETACHED=` immediately followed by the desired true/false value.
+* You can check your current value of `IS_DETACHED` at any time by typing `echo %IS_DETACHED%` in the command prompt.
+
+Linux
+---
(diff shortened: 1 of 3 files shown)


lib/sourceParams.js
```diff
@@ -127,9 +127,14 @@ module.exports = function (app) {
     params.htmlValidator.enable = false
   }
   if (flags.backgroundValidator) {
-    params.htmlValidator.separateProcess.enable = true
+    params.htmlValidator.separateProcess = true
```
Autre31415: `separateProcess` was made into an object when the autoKiller was added, so ensure you're interacting with the `enable` property. (This is why the build failed epicly)

JoshL18: Gotcha, I have added the `enable` property back to the `separateProcess` object

README.md
```diff
@@ -198,6 +198,33 @@ npm run audit
...
 
+Setting Up Environment Variable
+===
+
+Roosevelt's HTML Validator can be set up to be ran in the detached or attached state by default through use of the environment variable, `IS_DETACHED`.
+There are a few different ways that you can go about setting up this environment variable, depending on what OS your machine is running:
+
+Windows
+---
+1. Open up your Control Panel and navigate to System and Security, and then System.
```
alallier: I love the effort here, you did a great job. But it is way to verbose. Explaining how to set ENV variables is out of the scope of our project. At most we could link guides in how to do set ENV variables. That renders this section kinda outdated. We could create a section outlining all ENV variables, **and**/or add the ENV variable specifically to a section talking about the validator

lib/sourceParams.js
```diff
@@ -131,6 +131,11 @@ module.exports = function (app) {
   } else if (flags.attachValidator) {
     params.htmlValidator.separateProcess.enable = false
   }
+  if (process.env.IS_DETACHED === 'true') {
```
alallier: @JoshL18 we talked about this offline but I like the idea of namespacing ENV to the project, like `ROOSEVELT_IS_DETACHED`. Others are welcome to comment on this

lib/sourceParams.js
```diff
@@ -131,6 +131,11 @@ module.exports = function (app) {
...
+    params.htmlValidator.separateProcess.enable = true
```
alallier: I'm hitting this with a ENV set but it is not detaching for whatever reason

JoshL18:
I have opened up a [Pull Request](https://github.com/rooseveltframework/generator-roosevelt/pull/142) in the roosevelt-generator repository that this current PR (#472) depends on to work.

The PR I have linked to above is important for getting the ENV variable to work, as it will template the package.json with an updated version of the declaration of `separateProcess` that will enable the ENV variable to work as intended (declaring it as an object with properties rather than just a boolean)

README.md
```diff
@@ -177,6 +177,11 @@ npm run dev -- -attach-validator
 ```
 - `-a` can also be used.
 
+Roosevelt makes use of an ENV variable to set the validator to an attached or detached state by default. 
+- Set an environment variable `ROOSEVELT_IS_DETACHED` to true for a detached by default validator.
```
kethinov: change this to `ROOSEVELT_VALIDATOR` with values "detached" or "attached"

note that when this env var is set, it overrides whatever setting is in package.json or params

Autre31415: Should the env override the CLI as well?

kethinov: the hierarchy should be this:

- package.json
- params override package.json
- env vars override params
- CLI overrides env vars

might takes some sourceParams refactoring to make it behave this way iirc

alallier: ENV need higher priority. If you want to make params have priority over that's fine but ENV have to be over package.json

kethinov: i think you misread that. it says env should have higher priority over params

as for CLI vs env, CLI should take higher priority because its purpose is to override anywhere else is set, package.json, app params, etc... that includes env

kethinov: think of it like this: CLI is used to force things. if the user sets something in their CLI command, they *really mean it*. so it should take priority over everything

kethinov: @Autre31415 do we need to refactor the sourceParams stuff to get it to override in the above hierarchy? if so, i'd like to do it as part of this PR

JoshL18: @kethinov on my end it is currently working in the hierarchy you desire. the env variable will attach/detach as it's value states but will be overridden by CLI.

For example, having `ROOSEVELT_VALIDATOR` set to `detached`, then running the command `npm run dev -- -a` will force the `detached` state set by the ENV var to be overridden by the `-a` flag.

Issue #483, opened 2018-06-28
Issue with graceful shutdown when multi-threading


mulligan121:
When the --cores command line flag is in place and you try and do a graceful shutdown with `CTRL` + `C` it throws an error on Windows and Ubuntu. ~~~It seems to work fine on Mac though~~~ After further testing I do replicate this on Mac as well. Although I will note that the shutdown still works.

Ubuntu:
![cores2](https://user-images.githubusercontent.com/3014282/42044980-665955ae-7ac8-11e8-9159-3096e693f6d5.png)

Windows:
![cores2](https://user-images.githubusercontent.com/3014282/42045303-35e4c6aa-7ac9-11e8-85ba-e4a9c6a7035b.PNG)

Autre31415:
Good find, and good work fixing it!

PR #484, opened 2018-07-02
Fixes graceful shutdown when multi-threading

mulligan121:
This Close's #483 by fixing graceful shutdown to only store worker keys if graceful shutdown is called by the cluster master.


Autre31415:
Good find, and good work fixing it!

PR #496, opened 2018-07-18
Source files from .gitignore for our jsCompiler (closes #39)

### LINKED ISSUE #39: Files to ignore in jsCompiler.js should be sourced from .gitignore
Right now it's hardcoded to Thumbs.db.

JoshL18:
Now reads from our `.gitignore` file for `jsCompiler.js` instead of just looking for a single dot, two dots, or the `Thumbs.db` file

### DIFF
--- a/lib/jsCompiler.js
+++ b/lib/jsCompiler.js
@@ -19,6 +19,7 @@ module.exports = function (app, callback) {
   const fsr = require('./tools/fsr')(app)
   let preprocessor
   let jsFiles
+  let gitignoreFiles = []
   let preprocessorModule
   let preprocessorArgs
   let promises = []
@@ -84,6 +85,21 @@ module.exports = function (app, callback) {
     jsFiles = klawSync(jsPath)
   }
 
+  // read files to ignore from .gitignore
+  fs.readFile('.gitignore', (err, data) => {
+    let split
+    if (err) {
+      logger.error(err)
+    }
+    split = data.toString().split('\n')
+    split.forEach((line) => {
+      // skip blank lines
+      if (line !== '') {
+        gitignoreFiles.push(line)
+      }
+    })
+  })
+
   jsFiles.forEach((file) => {
     file = file.path || file
     promises.push(
@@ -105,7 +121,7 @@ module.exports = function (app, callback) {
           }
         }
 
-        if (file === '.' || file === '..' || file === 'Thumbs.db' || fs.lstatSync(usingWhitelist ? path.join(jsPath, file) : file).isDirectory()) {
+        if (gitignoreFiles.includes(file) || fs.lstatSync(usingWhitelist ? path.join(jsPath, file) : file).isDirectory()) {
           resolve()
         }
 


lib/jsCompiler.js
```diff
@@ -84,6 +85,21 @@ module.exports = function (app, callback) {
...
+  // read files to ignore from .gitignore
+  fs.readFile('.gitignore', (err, data) => {
+    let split
+    if (err) {
+      logger.error(err)
+    }
+    split = data.toString().split('\n')
+    split.forEach((line) => {
+      // skip blank lines
+      if (line !== '') {
```
kethinov: should account for other kinds of invalid entries too, like comments or whole lines filled with long whitespace

lib/jsCompiler.js
```diff
@@ -84,6 +87,13 @@ module.exports = function (app, callback) {
     jsFiles = klawSync(jsPath)
   }
 
+  if (fsr.fileExists(path.join(app.get('appDir'), '.gitignore'))) {
+    ignoreExists = true
+  } else {
+    logger.log('🔔', 'No .gitignore file found. Compiler will get files to ignore from list of defaults.'.yellow)
```
Autre31415: We don't need to log this stuff

lib/jsCompiler.js
```diff
@@ -105,8 +115,20 @@ module.exports = function (app, callback) {
           }
         }
 
-        if (file === '.' || file === '..' || file === 'Thumbs.db' || fs.lstatSync(usingWhitelist ? path.join(jsPath, file) : file).isDirectory()) {
-          resolve()
+        if (ignoreExists) {
+          if (dotgit.ignore(file) || file === '.' || file === '..') {
+            logger.log('💭', `${file} read from .gitignore and will not be compiled`)
```
Autre31415: No need for this log either

lib/jsCompiler.js
```diff
@@ -105,8 +115,20 @@ module.exports = function (app, callback) {
...
+            resolve()
+          }
+        } else {
+          let length = defaults.length
+          // scan through list of defaults for file
+          while (length--) {
+            if (file === '.' || file === '..' || file.indexOf(defaults[length]) !== -1 || fs.lstatSync(usingWhitelist ? path.join(jsPath, file) : file).isDirectory()) {
+              logger.log('💭', `${file} read from defaults and will not be compiled`)
```
Autre31415: Or this log either

package.json
```diff
@@ -19,6 +19,7 @@
     "colors": "~1.3.0",
     "compression": "~1.7.0",
     "cookie-parser": "~1.4.0",
+    "dotgitignore": "~1.0.3",
```
kethinov: adding a whole dependency to split a file's contents by newlines isn't really necessary

JoshL18: this dependency is also doing the checking for us whether or not a file is affected by the .gitignore, returning true or false whether or not the file you pass is it meant to be ignored by gitignore. this saves us having to do a lot of things besides just splitting a file into new lines

kethinov: that's not hard to do manually though and the dependency isn't a terribly active project. we've run into problems in the past with adding dependencies that aren't being actively maintained and were never much code to begin with. the rule of thumb is a dependency needs to have some decent level of substance and active maintenance behind it before it's worth including in a project, and this one isn't quite passing the sniff test

lib/jsCompiler.js
```diff
@@ -18,6 +19,7 @@ module.exports = function (app, callback) {
   const logger = require('./tools/logger')(app.get('params').logging)
   const fsr = require('./tools/fsr')(app)
   let preprocessor
+  let gitignoreFiles = gitignoreScanner(path.join(app.get('appDir'), '.gitignore'))
```
Autre31415: Don't run the gitignore scanner until just before it's needed, that way it only gets run when the compiler is enabled and ready to go.

lib/tools/gitignoreScanner.js
```diff
@@ -0,0 +1,35 @@
...
+  } catch (e) {
+    // if we can not, return list of defaults
+    return fileList
+  }
+
+  // get data from .gitignore, trim it of whitespace, split at new lines
+  gitData = fs.readFileSync(gitignorePath).toString().trim().split('\n')
+
+  // skip lines already in fileList as well as blank lines/comments
+  gitData.forEach((line) => {
```
Autre31415: Trim in individual line in this loop before examining the contents to defend against random whitespace.

lib/tools/gitignoreScanner.js
```diff
@@ -0,0 +1,41 @@
+// Module for scanning the .gitignore and constructing a list of files to ignore
+const fs = require('fs')
+const path = require('path')
+
+module.exports = function (gitignorePath) {
+  let fileList = ['lib-cov', 'node_modules', 'pids', 'logs', 'results', 'Thumbs.db', '*.seed', 'npm-debug.log', '.build', 'public', '*.log', '*.csv', '*.dat', '*.out', '*.pid', '*.gz', '.DS_Store']
```
kethinov: the ones with * in them can go

lib/tools/gitignoreScanner.js
```diff
@@ -0,0 +1,41 @@
...
+  try {
+    // ensure that we can access .gitignore
+    fs.accessSync(gitignorePath)
+  } catch (e) {
+    // if we can not, return list of defaults
+    return fileList
+  }
+
+  // get data from .gitignore, trim it of whitespace, split at new lines
+  gitData = fs.readFileSync(gitignorePath).toString().split('\n')
```
kethinov: 1. why is `toString()` necessary? shouldn't `readFileSync` give you a string already?
2. does this account for differences in how each OS handles line endings? what if the .gitignore file is authored in a windows environment?

lib/tools/gitignoreScanner.js
```diff
@@ -0,0 +1,41 @@
...
+
+  // skip lines already in fileList as well as blank lines/comments
+  gitData.forEach((line) => {
+    line = line.trim()
+    if (!fileList.includes(line)) {
+      if (line !== '' && !line.startsWith('#') && !line.endsWith('.js') && !line.endsWith('.css') && !line.endsWith('.less') && !line.endsWith('.sass') && !line.endsWith('.scss') && !line.endsWith('.jsx')) {
```
kethinov: throw in .coffee, .litcoffee, .ts, and .tsx

lib/tools/gitignoreScanner.js
```diff
@@ -0,0 +1,41 @@
...
+        fileList.push(line)
+      }
+    }
+  })
+
+  // apply basename to each line
+  fileList.forEach((file) => {
+    if (file.startsWith('*')) {
+      let ext = file.split('*')
```
kethinov: straight up discard any line with * in it

lib/tools/gitignoreScanner.js
Autre31415: For point 2 I think it would be safer to use `.split(/\r?\n/)` to ensure OS compatibility.

JoshL18: For point 1, just using `readFileSync` returns a buffer string to gitData like:
 `<Buffer 6c 69 62 2d 63 6f 76 0a 2a 2e 73 65 65 64 0a 2a 2e 6c 6f 67 0a 2a 2e 63 73 76 0a 2a 2e 64 61 74 0a 2a 2e 6f 75 74 0a 2a 2e 70 69 64 0a 2a 2e 67 7a 0a ... >`
 Then after appending a `toString()` to it, gitData will get returned the content of the gitignore file like we need

lib/tools/gitignoreScanner.js
```diff
@@ -0,0 +1,40 @@
...
+    if (!fileList.includes(line)) {
+      if (line !== '' && !line.startsWith('#') && !line.endsWith('.js') && !line.endsWith('.css') && !line.endsWith('.less') && !line.endsWith('.sass') && !line.endsWith('.scss') && !line.endsWith('.jsx') && !line.endsWith('.coffee') && !line.endsWith('.litcoffee') && !line.endsWith('.ts') && !line.endsWith('.tsx')) {
+        fileList.push(line)
+      }
+    }
+  })
+
+  // apply basename to each line
+  fileList.forEach((file) => {
+    if (file.startsWith('*')) {
```
kethinov: if the line contains * anywhere, we should toss it. not just starts with

PR #498, opened 2018-07-20
Fix ECMA Deprecation Warning

mulligan121:
Currently when running lint there's a deprecation warning for `ecmaFeatures.experimentalObjectRestSpread`

capture

This is because that syntax is now supported in ECMA 2018. A link to this documentation is here https://eslint.org/docs/user-guide/configuring#specifying-parser-options

This PR just updates to ECMA 2018 and removes ecmaFeatures from package.json.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4633,7 +4633,6 @@
           "version": "0.1.4",
           "bundled": true,
           "dev": true,
-          "optional": true,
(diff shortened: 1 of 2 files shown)


Autre31415:
Could you eliminate the package-lock.json changes from this PR?

PR #503, opened 2018-07-24
Added param for --cores

JoshL18:
Now lets us activate and tweak multi threading support via source param `cores`

### DIFF
--- a/test/util/testConstructorConfig.json
+++ b/test/util/testConstructorConfig.json
@@ -69,5 +69,5 @@
   "favicon": "constructor",
   "versionedPublic": "constructor",
   "alwaysHostPublic": "constructor",
-  "cores": 1
+  "cores": "constructor"
 }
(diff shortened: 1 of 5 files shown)


test/util/testConstructorConfig.json
```diff
@@ -68,5 +68,6 @@
   "publicFolder": "constructor",
   "favicon": "constructor",
   "versionedPublic": "constructor",
-  "alwaysHostPublic": "constructor"
+  "alwaysHostPublic": "constructor",
+  "cores": 1
```
Autre31415: Set this to "constructor" like the other items. It looks silly, but all we're testing for here is that this param value is what gets set.

test/util/testPkgConfig.json
```diff
@@ -72,5 +72,6 @@
   "publicFolder": "package",
   "favicon": "package",
   "versionedPublic": "package",
-  "alwaysHostPublic": "package"
+  "alwaysHostPublic": "package",
+  "cores": 1
```
Autre31415: Set this one to "package" like the others as well

Autre31415:
LGTM

PR #505, opened 2018-08-02
Added a syntax highlighter for validator error page

### LINKED ISSUE #100: HTML validator should display error beside line number

JoshL18:
(closes #100, #101)
Now displays our invalid HTML with highlighted syntax along with line numbers on the side.
This change includes the use of the node module Prism.js syntax highlighter to apply HTML syntax highlighting to our htmlValidator's error page.
Also, highlights the line number in red that the error occured on. A user can `HOVER` the line number, and the page will display what error is on that line.

Here is an example of how any invalid html will be displayed to the user (didn't close article tag):
image

Here is a video of the `HOVER` in action:
highlighter


roosevelt.js
```diff
@@ -39,6 +39,7 @@ module.exports = function (params) {
   // expose initial vars
   app.set('express', express)
   app.set('params', params)
+  app.use('/scripts', express.static(params.appDir + '/node_modules/prismjs/'))
```
Autre31415: This should be set in htmlValidator js in the same section where the page template is being pieced together.

PR #507, opened 2018-08-03
Scanner for detecting stale files in .build

### LINKED ISSUE #282: Scanner to detect stale files in .build and recommend running clean
Something like "hey there are files that are getting quite stale in your `.build`, maybe run npm run clean?"

Determine this by checking last modified dates relative to other files in `.build`. Perhaps anything older than x days triggers the warning?

JoshL18:
(closes #282)
Added buildScanner module that reports when stale files are in user configured .build directory


lib/tools/buildScanner.js
```diff
@@ -0,0 +1,43 @@
+// Module for scanning the .build directory and alerting the user when they have stale files
+const fs = require('fs')
+const path = require('path')
+
+module.exports = function (buildPath) {
+  let buildFiles
+  let stale = false
+
+  var walkSync = function (dir, filelist) {
```
Autre31415: We're already using klawSync, which is a module that does exactly this.

roosevelt.js
```diff
@@ -62,6 +64,10 @@ module.exports = function (params) {
 
   logger.log('💭', `Starting ${appName} in ${appEnv} mode...`.bold)
 
+  haveStale = staleCheck()
```
Autre31415: To simplify this workflow a bit how about moving all of the functionality of the check into the buildScanner module (Including the logging part). Then all roosevelt.js needs to do is call it.

roosevelt.js
```diff
@@ -62,7 +63,8 @@ module.exports = function (params) {
 
   logger.log('💭', `Starting ${appName} in ${appEnv} mode...`.bold)
 
-  // let's try setting up the servers with user-supplied params
+  buildScanner(app)
```
kethinov: i think we should move this later in the execution somewhere after compileJs so that old files have a chance to be overwritten (and thus their dates updated) first

lib/tools/buildScanner.js
```diff
@@ -0,0 +1,37 @@
...
+    // make sure we can access the .build directory
+    fs.accessSync(buildPath)
+  } catch (e) {
+    return
+  }
+
+  buildFiles = klawSync(buildPath, {nodir: true})
+
+  // find the date for each of these files and compare to current date
+  // if the difference is greater than 604800000ms (a week), we will warn the user about a stale file.
```
kethinov: make this configurable

lib/tools/buildScanner.js
```diff
@@ -0,0 +1,37 @@
...
+    let dateAccessed = new Date(fs.statSync(file).mtimeMs)
+    let currentTime = new Date().getTime()
+    let date = new Date(currentTime)
+    let difference = date - dateAccessed
+    if (difference >= 604800000) {
+      stale = true
+    }
+  })
+  if (stale) {
+    logger.warn(`You have stale file(s) in your ${compDir} directory. You may want to run npm run clean.`)
```
kethinov: change to `There are stale file(s) in ${compDir}. You may want to run npm run clean.`

also show the list of stale files

README.md
```diff
@@ -370,6 +370,13 @@ App behavior parameters
 - `cores`: Parameter to activate multi threading support on roosevelt. By default, roosevelt will run on 1 core, but you can change the number of cores that the app will run on with this param. To use all available cores, set this value to `max`.
   - Default: *[Number]* `1`
 
+- `staleTimer`: Parameter to set time (in `milliseconds`) a file must reside in the user's compiled JS directory before the `buildScanner` tool reports that it is stale.
```
kethinov: - this belongs in the statics section
- rename param to `cleanTimer`
- if set to to `0`, `null`, or anything that isn't a number, disable the check entirely.

rewrite to description to:

- Time in milliseconds to allow before considering files in CSS/JS compile directories stale and recommending running `npm run clean`.
  - Default: *[Number]* `604800000` (1 week)
  - Useful time conversions to `milliseconds` to configure this param with:
    - `1 day`: `86400000`
    - `1 week`: `604800000`
    - `1 month`: `2419200000`
  - Set to `0`, `null`, or anything that isn't a number to disable the check entirely.

roosevelt.js
```diff
@@ -182,6 +182,8 @@ module.exports = function (params) {
       require('./lib/htmlValidator')(app, mapRoutes)
     }
 
+    buildScanner(app)
```
kethinov: this will not execute in the order you expect

look at the code preceding this. each function calls the next

you'll want to follow the pattern. define a function and call it from `validateHTML`, then have your function call `mapRoutes` the way `validateHTML` was

lib/tools/buildScanner.js
```diff
@@ -0,0 +1,40 @@
+// Module for scanning the .build directory and alerting the user when they have stale files
+require('colors')
+const fs = require('fs')
+const path = require('path')
+const klawSync = require('klaw-sync')
+
+module.exports = function (app) {
+  const logger = require('./logger')(app.get('params').logging)
+  let buildFiles
+  let compDir = app.get('params').js.output.split('/')[0]
```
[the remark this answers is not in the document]
JoshL18: The scanner does technically account for built css as well as `app.get('params').js.output` returns the string `.build/js` and then the `split('/')[0]` appended to the end of the statement gives us `.build`. 
I did this to avoid hardcoding `.build` in there as it is configurable.
I just wanted to let you know that, so I can clear up exactly what is needed. I could proceed by writing a similar statement such as `app.get('params').css.output.split('/')[0]`, which would be needed in the case that the user has different `.build` directories for their css and js, if you think considering for that case is necessary.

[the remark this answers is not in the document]
JoshL18: Scratch what I had said before, the solution I just pushed up should be what you want.

kethinov:
also the new unit test appears to be failing

PR #512, opened 2018-08-09
Winston Logger

mulligan121:
This PR makes a few changes to the logging system that Roosevelt currently has.

- [x]  Use Winston for the main logger in Roosevelt
- [x]  Expose the Roosevelt logger & Winston
   1.  `app.get('logger')` will include the Roosevelt logger which can be configured by parameters in `package.json`
   2.  `app.get('winston')` will include the winston module itself
- [x]  Make parameters more customizable
```js
"logging": {
  "http": true,
  "appStatus": true,
  "warnings": true,
  "verbose": true,
  "logType1": true,
  "logType2": {
    "type": "warn"
  },
  "logType3": {
    "enabled": true
  },
  "logType4": {
    "enabled": true,
    "type": "info"
  },
  "logType5": false
}
```
Custom log types can now be created from parameters, for example if a user wants to add a debugging log type they would just need to add:
```js
"logging": {
  "debug": true,
  // or
  "debug": {
    // enabled is optional (defaults to true if omitted)
    "enabled": true
     // types can be set (ex: info, warn, error) (defaults to 'info' if omitted)
    "type": "error"
  }
}
```
To use the debug log type: 
`logger.debug('Hello World')`
- [x]  Make the JSON log data look nicer with [util.inspect()](https://nodejs.org/api/util.html#util_util_inspect_object_options)
- [x]  Allow for different environments to suppress logs
To suppress logs in certain environments you can now add `disable` to the `logging` parameter. `disable` takes an array of items to suppress, for example:
if you're in a mocha process
`process.env.LOADED_MOCHA_OPTS = true`
or if you're in production
`process.env.NODE_ENV = 'production'`
so you would set these as items in the array
```
"logging": {
  disable: ['LOADED_MOCHA_OPTS', 'production']
}
```
This also keeps the spacing consistent with multiple arguments as we had before.


roosevelt.js
```diff
@@ -220,7 +225,7 @@ module.exports = function (params) {
     }, app.get('params').shutdownTimeout)
 
     app.set('roosevelt:state', 'disconnecting')
-    logger.log('\n💭 ', `${appName} received kill signal, attempting to shut down gracefully.`.magenta)
```
Autre31415: I know that `\n` looks silly, but it was there so the log wouldn't look like this:

silly log

README.md
```diff
@@ -299,6 +300,17 @@ App behavior parameters
         "verbose": false
       }
       ```
+  - Custom: *[Object]*
+      ```json
+      {
+        "debug": {
+          // enabled is optional (defaults to true if omitted)
+          "enabled": true,
```
Autre31415: Rename this to `enable` to match other similar params.

Autre31415:
Some issues:

- I created a custom log called debug like this:
  ```
  "debug": {
    "enabled": true,
    "type": "info"
  }
  ```
  and `logger.debug` does work, but when I set `enabled` to false I still see the log.

- When I create that example and run the config auditor it complains about the extra param:
  `⚠️  Extra param "debug" found in "logging", this can be removed.`

PR #539, opened 2018-10-03
Adding option to expose view templates to frontend JS

### LINKED ISSUE #298: API for scooping up views and exposing them to frontend JS as packaged templates
- Off by default
- When enabled:
  - Defaults to whitelisting
  - Should have a whitelist param like:
    ```javascript
    {
      "bundles": {
        "path/to/generated/js/file.js": [ /* list of template files */ ],
        "path/to/other/generated/js/file.js": [ /* list of template files */ ],
        etc...
      }
    }
    ```
  - Should also have a blacklisting param to list templates to always exclude
  - Off by default option to scoop up all templates except those on the blacklist
  - Templates can declare themselves for a bundle Inline at the top line of a given template's source via a code comment containing a special syntax indicating that the template is either whitelisted or blacklisted; this line should be removed once the template is packaged for frontend delivery
    - If a template doesn't declare which bundle it wants to go in, it goes in the first bundle
  - Option to supply middleware to preprocess templates prior to packaging for frontend delivery (useful for some templating systems, e.g. for teddy a dev can optionally run it through a `teddy.compile` before sending to the frontend) 
  - Option to minify template source prior to packaging for frontend delivery (on by default?)
  - Add this comment to the top of all generated JS bundles: `/* do not edit; generated automatically by Roosevelt */`

lannonbr:
This is going to resolve #298. I will continue on working on enhancing this to include all of the features requested in said issue

Features Requested:

- [x] Exposing off by default
- [x] Defaulted to whitelisting
- [x] bundle map syntax
- [x] blacklisting param
- [x] option to expose all templates but ones on blacklist
- [x] declare where a bundle goes with syntax at top of template
- [x] Supply preprocess middleware option
- [x] minify template source before pushing to frontend
- [x] add comment at top of JS bundle that it was generated and shouldn't be edited
- [x] add error handling across the module


lib/sourceParams.js
```diff
@@ -162,6 +163,9 @@ module.exports = function (app) {
   app.set('jsCompiledOutput', path.join(appDir, params.staticsRoot, params.js.output))
   app.set('jsBundledOutput', path.join(app.get('jsPath'), params.js.bundler.output))
 
+  // map shared template files
+  app.set('clientViewBundles', params.clientViewBundles)
```
Autre31415: Since you're exposing this as an express variable, be sure to add it to the [Express variables exposed by Roosevelt table](https://github.com/rooseveltframework/roosevelt/blob/f0c6765e7ebbbc927d833b5bcc69e4d300bc1c48/README.md#express-variables-exposed-by-roosevelt)

lib/tools/viewsBundler.js
```diff
@@ -0,0 +1,75 @@
...
+    } catch (e) {
+      console.log('📁  ' + ('Roosevelt making new directory ' + basePath).warn)
+      fse.ensureDirSync(basePath)
+    }
+
+    fileData = '/* Do not edit; generated automatically by Roosevelt */\n\n/* eslint-disable */\n\n'
+
+    fileData += 'module.exports = function() {\n  return ' + JSON.stringify(templates) + '\n}\n'
+
+    console.log('📝' + '  ' + (appName + ' writing new JS file ' + writePath).success)
```
Autre31415: Couple points here:
- Use Roosevelt's internal logger for these status messages.
- Since this is writing files, the `generateFolderStructure` param needs to be able to prevent it, the built-in `fsr` module integrates logic for this. (This should be the case anywhere a file/directory would be written)

[See here for an example](https://github.com/rooseveltframework/roosevelt/blob/master/lib/jsCompiler.js#L145)

lib/sourceParams.js
lannonbr: Done

lib/tools/viewsBundler.js
lannonbr: Done

README.md
```diff
@@ -676,6 +676,15 @@ Statics parameters
...
       ```
 
+- `clientViewBundles`: *[Object]* A map which will expose templates to clientside JS. Each entry is an array of templates which will be saved in the module defined in the key. The base path for where the client exposed templates are saved is `staticsRoot/.build/templates`
+  - Example:
+  ```json
+  {
+    "mainLayouts.js": ["baseLayout.html", "footer.html"],
+    "forms.js": ["forms/formTemplate.html"],
+  }
+  ```
```
kethinov: show the example source code of one of these files

lib/tools/viewsBundler.js
```diff
@@ -0,0 +1,76 @@
...
+  let fileNameAndPath
+  let fileName
+  let contents
+  let appName = app.get('appName')
+  let appDir = app.get('appDir')
+  let viewsPath = app.get('viewsPath')
+  let staticsRoot = app.get('staticsRoot')
+  let bundles = app.get('clientViewBundles')
+  let bundlesLength = Object.keys(bundles).length
+  let basePath = path.join(appDir, staticsRoot, '.build/templates')
```
kethinov: this path should be configurable

lannonbr: I'm thinking of doing it similar to how the JS / CSS which is bundled up for the frontend has an output path that is specified in the configs.

lannonbr:
Question, what should have higher priority? A tag defined in the templates on a template being whitelisted / blacklisted or the Roosevelt config stating a file is whitelisted / blacklisted?

lannonbr:
My particular views on the above comment on precedence is as follows for each instance:

1) I think blacklisting should overrule whitelisting so if there is a `<!-- roosevelt-blacklist -->` tag, it will not be exposed
2) Similar to the reasoning of the first, having a template in the blacklist param in the Roosevelt config should not expose it even if it has a `<!-- roosevelt-whitelist -->` tag.
3) I do not have a particular say with this instance

lannonbr:
After talking with @Autre31415 on this, the blacklisting taking precedence is a good solution since it does similar with the JS compiler does similar of blacklists > whitelists.

As well, given the tags are more brute force in a way, it should take precedence in issue 3.

Finally, All of these instances should have the logger send a warning stating that both syntaxes are being used and which one is taking precedence.

kethinov:
Agreed. Pick a precedence, make it consistent, declare the precedence in the README somewhere, and show a logger warning if there are conflicting declarations like the scenarios you described.

kethinov:
closed view https://github.com/rooseveltframework/roosevelt/pull/824

Issue #380, opened 2018-02-26
Audit tests to be more concise if possible


Autre31415:
The generateTestApp pattern is used in a majority of the unit tests so far. We should audit this usage and simplify it anywhere we can.

Autre31415:
https://github.com/rooseveltframework/roosevelt/pull/409 is a sample of how we might accomplish this

Percipient24:
So this is going to be verbose... My apologies!

I chose to start poking at the jsBundler first, because it wasn't too short, wasn't too long, had some `require`s in there, and used `generateTestApp` in its existing tests.

In looking over the unit tests, the app has good coverage:
```
13 minutes for 342 tests (~2.28 seconds per test)
-----------------------------------------|----------|----------|----------|----------|-------------------|
File                                     |  % Stmts | % Branch |  % Funcs |  % Lines | Uncovered Line #s |
-----------------------------------------|----------|----------|----------|----------|-------------------|
All files                                |    99.58 |    97.86 |    98.64 |    99.75 |                   |
 roosevelt                               |    98.85 |    96.23 |      100 |    98.85 |                   |
  roosevelt.js                           |    98.85 |    96.23 |      100 |    98.85 |           301,302 |
 roosevelt/defaultErrorPages/controllers |      100 |      100 |      100 |      100 |                   |
  404.js                                 |      100 |      100 |      100 |      100 |                   |
  503.js                                 |      100 |      100 |      100 |      100 |                   |
  5xx.js                                 |      100 |      100 |      100 |      100 |                   |
 roosevelt/lib                           |    99.59 |    97.85 |    97.44 |    99.86 |                   |
  500ErrorPage.js                        |      100 |      100 |      100 |      100 |                   |
  enableMultipart.js                     |      100 |      100 |      100 |      100 |                   |
  htmlMinify.js                          |      100 |      100 |      100 |      100 |                   |
  htmlValidator.js                       |      100 |      100 |      100 |      100 |                   |
  jsBundler.js                           |       98 |       70 |      100 |       98 |                42 |
  jsCompiler.js                          |      100 |      100 |      100 |      100 |                   |
  mapRoutes.js                           |     97.8 |      100 |       75 |      100 |                   |
  preprocessCss.js                       |      100 |      100 |      100 |      100 |                   |
  setExpressConfigs.js                   |      100 |      100 |      100 |      100 |                   |
  sourceFlags.js                         |      100 |      100 |      100 |      100 |                   |
[...]

PR #562, opened 2018-10-29
Revised Unit Tests For jsBundler.js

Percipient24:
**tl;dr** - Here are some unit test changes. Do we like this testing approach? Should I continue with all the remaining tests?

----

As discussed in #380 (and offline) I went ahead and:

1. Renamed `/test/unit` to be `/test/integration`
1. Created a new `/test/unit` directory, and started building it out with a test-file-per-module.
1. Made the `proxyquire` and `app` mocks faithful, and moved them to external files for re-usability (see `test/util/mocks/*`.
1. Refactored `jsBundler.js` itself to be a little more modular.
1. Wrote new unit tests for jsBundler, following this new approach.

As a result:

1. Unit tests run faster (jsBundler now runs 45 tests in 3 seconds, compared to 12 tests in 19 seconds).
1. Unit tests are more granular (see output below), and we will have minimal phantom coverage of one module on another (see coverage below that).
1. `jsBundler.js` itself is much easier to follow, reason about, and test.

There are some things I'm not thrilled with, and I'm open to suggestions on fixing:

1. Excessive use of `this.` in jsBundler - though I think that's a necessary part of chunking out the logic into smaller methods.
1. jsBundler now exposes its references to `fsr`, `logger`, `path`, and `browserify` (and all of the `params` etc it lifts off of `app`). This is a partial side effect of testing, and a partial side effect of the mocks/modules themselves being factories.

There are some things I noticed during the testing:

1. I'm not sure the existing (now integration) tests clean up after themselves well. Case in point (see inline comment) when the shape of `jsBundler` changed, a ::lot:: of the `jsBundler` tests continued to pass. It took me longer than I'd care to admit to realize the first eight or so tests were giving false positives.
1. There are three cases where making directories / files could fail and it would fail silently. (See `fsr.ensureDirSync` and `fsr.writeFileSync`) Should I add an else with an error (or am I missing a subtlety)?
1. `proxyquire` is awesome. :-)

```
lib/jsBundler
    .bundle
      when bundler is disabled
        √ should call the provided callback
        √ should not continue with the operation
      when bundler is enabled
        √ should check the output directory
        √ should check the build directory
        √ should check each bundle for inclusion
        when the bundle should be included
          √ should process the bundle
          √ should call conclude with the proper number of promises
        when the bundle should be omitted
          √ should not process the bundle
          √ should call conclude with the proper number of promises
    .computeBuildPath
      √ should return a build path
    .isBundlerEnabled
      √ should return true when params.js.bundler.bundles.length has a nonzero length
      √ should return false when params.js.bundler.bundles.length has a zero length
    .checkOutputDir
      when the output directory exists
        √ should do nothing
      when the output directory does not exist
        √ should make the directory, and log the result (assumes success)
    .checkBuildDir
      when expose is false
        √ should do nothing, even if the directory exists
      when expose is true
        √ should do nothing if the directory exists
        √ should make the directory, and log the result (assumes success)
    .shouldIncludeBundle
      during a dev build
        √ should return true for a dev bundle
        √ should return true for a non-dev bundle
      during a prod build
        √ should return false for a dev bundle
        √ should return true for a non-dev bundle
    .processBundle
      √ should preprocess the bundle files
      √ should run browserify with the bundle
      √ should run browserify.bundle with a function
      √ should handle the browserify result
      √ should return a promise that can resolve
      √ should return a promise that can reject
    .processBundleFiles
      √ should pathify the bundle's files
      √ should pathify the bundle's paths
      √ should provide a default path if the bundle has no params
    handleBrowserifyResult
      when there is an error
        √ should log an error message
        √ should reject the promise
        √ should should not continue with the operation
      when there is no error
        when expose is false
          √ should write to the output dir and log
          √ should not write to the build dir
          √ should resolve the promise
        when expose is true
          √ should write to the output dir and log
          √ should write to the build dir and log
          √ should resolve the promise
    .conclude
      when promises all resolve
        √ should call the callback
        √ should not log errors
        √ should not exit the process
      when a promise rejects
        √ should not call the callback
        √ should log the error
        √ should exit the process with code 1


  45 passing (3s)

---------------|----------|----------|----------|----------|-------------------|
File           |  % Stmts | % Branch |  % Funcs |  % Lines | Uncovered Line #s |
---------------|----------|----------|----------|----------|-------------------|
All files      |     49.3 |    28.89 |    48.39 |       50 |                   |
 lib           |      100 |    86.67 |      100 |      100 |                   |
  jsBundler.js |      100 |    86.67 |      100 |      100 |     57,66,121,132 |
 lib/tools     |    13.25 |        0 |        0 |    13.58 |                   |
  fsr.js       |    13.04 |        0 |        0 |    13.64 |... 39,44,45,46,48 |
  logger.js    |    13.33 |        0 |        0 |    13.56 |... 39,145,146,152 |
---------------|----------|----------|----------|----------|-------------------|
```


roosevelt.js
```diff
@@ -176,7 +176,7 @@ module.exports = function (params) {
     }
 
     function bundleJs () {
-      require('./lib/jsBundler')(app, compileJs)
+      require('./lib/jsBundler').bundle(app, compileJs)
```
Percipient24: Without this change, `npm run test:integration` still passes the first 8 or so tests for `jsBundler`. I'm not sure if a fresh run behaves that way, or if I had phantom files leftover from a previously successful run.

kethinov:
Closing this due to offline conversations + significant drift from the current state of the codebase.

Issue #520, opened 2018-08-30
Inconsistent behavior with alwaysHostPublic in production mode


alallier:
According to the docs:

> By default in production mode Roosevelt will not expose the public folder.

This is both true and false. If the application is start in prod mode by the ENV variable `alwaysHostPublic` will be off.

But if it is started in prod mode by specifying the prod flag `npm run --production-mode` `alwaysHostPublic` will be on due to these [lines](https://github.com/rooseveltframework/roosevelt/blob/master/lib/sourceParams.js#L140-L142)

So in some cases with production mode `alwaysHostPublic` if on and sometimes it is off.

This probably needs a unit test since we have 100% coverage

Issue #462, opened 2018-05-29
Document reserved CLI flags in one section in README


alallier:
Might be nice to document reserved/used CLI flags in the README in one section so that when developing an application you can easily know what to steer clear of. Right now this information is mostly available but it involves reading/troweling a good portion of the README.

Issue #381, opened 2018-02-26
Investigate the defaults we use with jsBundler


Autre31415:
By default `expose` is `true`.
- When enabled the bundled js file is copied to the public folder.

We might consider defaulting this to false, since after the bundle is created it will end up being compiled, so why have the uncompiled bundle in the public folder by default?

PR #567, opened 2018-11-05
WIP: README overhaul and other prerelease cleanup

kethinov:
closes https://github.com/rooseveltframework/roosevelt/issues/381

closes https://github.com/rooseveltframework/roosevelt/issues/462

closes https://github.com/rooseveltframework/roosevelt/issues/520


lib/sourceParams.js
```diff
@@ -35,13 +35,13 @@ module.exports = function (app) {
...
   // default env to production
   process.env.NODE_ENV = 'production'
+  app.set('env', 'production')
 
   pkg.rooseveltConfig = pkg.rooseveltConfig || {}
 
   app.set('appDir', appDir)
   app.set('package', pkg)
   app.set('appName', pkg.name || 'Roosevelt Express')
-  app.set('appVersion', pkg.version)
```
lannonbr: Removing this breaks a test in cssTest.js due to the fact that when you try to get the version code in the various preprocessors, they still depend on the appVersion field to be set.

test/unit/defaultParams.js
```diff
@@ -11,7 +11,7 @@ describe('Default Parameter Tests', function () {
   before(function () {
     app = require('../../roosevelt')({
       appDir: path.join(__dirname, '../app/defaultParams'),
-      ignoreCLIFlags: true,
+      enableCLIFlags: false,
```
lannonbr: When running through the test here for enableCLIFlags, it will say enableCLIFlags is false here when it is true in the defaults.json file. Is it supposed to be true or false by default?

lannonbr: This line can be removed and the tests will all still pass

lannonbr:
So as of 3bc05c0, the only test failures left are related to the https server options tests, but I thought those were resolved with PR #573. @techrev, do you have any possible insights?

Issue #559, opened 2018-10-25
Add support for prefixed routes


mulligan121:
Currently, roosevelt uses `klawSync` to walk the controllers directory to set up routing. It would be useful if we could specify certain or all controllers to have a prefixed route. This could be a configuration inside of the `rooseveltConfig`

```javascript
// for all controllers
urlPrefix: '/foo'
```

or

```javascript
urlPrefix: [{
  prefix: '/foo',
  controllers: ['controller1.js', 'controller2.js']
}, {
  prefix: '/bar',
  controllers: ['controller3.js', 'controller4.js']
}]
```

**Things to think about:**
- ~~Prefixed route controllers would need to be written differently because they would be an express router (`express.Router()`) and be separate from the main applications default router.~~ **EDIT: This turns out to be a false statement.**

- If we specify certain controllers to have a prefix, than we would need to account for duplicate controller names if they are nested in different directories.
  - Example, suppose `/foo` controllers are in a directory named foo and `/bar` controllers are in a directory named bar
    ```javascript
     urlPrefix: [{
       prefix: '/foo',
       controllers: ['controller1.js', 'controller2.js']
     }, {
       prefix: '/bar',
       controllers: ['controller1.js', 'controller4.js']
     }]
     ```
  - Maybe require relative or absolute paths to be passed starting in the controller directory ex:
    - `controllers: ['foo/controller1.js', 'foo/controller2.js']`
- Also we would probably want a way to programmatically determine if a controller is part of the main app router or a separate express router.

**Other thoughts:**
- What if for larger applications there could be a file in the root of the controller directory that specifies all prefixes and relative or absolute paths to those controller files? I think that package.json would get too big for a large application using many prefixed routes.

PR #570, opened 2018-11-06
Route prefix configuration

mulligan121:
This will allow for custom routers/static prefixes to be established through configs which include a path and and a set of files/directories to use with that router/static prefix. 

```json
"routers": {
  "controllers": [
    {
      "prefix": "/foo",
      "files": [
        "homepage.js",
        "subfolder"
      ]
    },
    {
      "prefix": "/bar",
      "files": [
        "bar/controller1.js",
        "bar/controller2.js"
      ]
    }
  ],
  "statics": [
    {
      "prefix": "/public",
      "files": [
        "css", 
        "images", 
        "js"
      ]
    }
  ]
}
```

- `routers` *[Object]* - The parent parameter object that accepts two keys, `controllers` & `statics`
- `controllers` *[Array]* - Array of objects that represent express routers (**Note:** A string can be supplied instead of an array. This will represent a file within the controllers directory that exports the array of objects)
  1. `prefix` *[String]* - The path supplied to express for the router
  2. `files` *[Array]* - An array of **files or directories** within the controllers directory that will be used with the express router
    - Controllers Example:
       - 
      ```javascript
      routers: {
        controllers: [
          {
            prefix: '/home',
            files: ['homepage.js', 'subfolder']
          }
        ]
      }
      ```
      - If `homepage.js` has a `GET` route `/` you would be able to access that at http://localhost:43711/home/
- `statics` *[Array]* - Array of objects that represent static file routes
  1. `prefix` *[String]* - The path supplied to express that will be the prefix for static directories 
  2. `files` *[Array]* - An array of **directories** within the statics directory that will have an added prefix.
    - Statics Example:
      - 
      ```javascript
      routers: {
        statics: [
          {
            prefix: '/public',
            files: ['css', 'images', 'js']
          }
        ]
      }
      ```
     - You will be able to access the static files from http://localhost:43711/public/css/style.css
---

Supports #559


lib/defaults/config.json
```diff
@@ -84,5 +84,6 @@
   "alwaysHostPublic": false,
   "checkDependencies": true,
   "shutdownTimeout": 30000,
-  "cores": 1
+  "cores": 1,
+  "urlPrefix": false
```
alallier: Shouldn't the default be `/`? Otherwise we have a mix of types between boolean and string

```suggestion
  "urlPrefix": "/"
```

lib/mapRoutes.js
```diff
@@ -177,3 +193,18 @@ module.exports = function (app) {
 
   return app
 }
+
+function validatePrefix (urlPrefix) {
```
alallier: Does this handle these three cases and always return the same prefix? I'd recommend expanding the unit tests a little more

* `prefix`
* `/prefix`
* `prefix`
* `/prefix/`
* `prefix/another`
* `/prefix/another`
* `prefix/another/`
* `/prefix/another/`

alallier: I imagine the context paths could be n levels deep as well.

`/prefix/another/more/.../`

mulligan121: Should be all set with those cases. It also works levels deep now. If there is no leading `/` passed such as the `prefix` case one will be appended to the first index of the string.

README.md
```diff
@@ -431,6 +431,11 @@ App behavior parameters
 
   - Default: *[Number]* `30000` (30 seconds).
 
+- `urlPrefix`: A prefix to append to all routes within the controller files.
```
kethinov:
```suggestion
- `urlPrefix`: A string to prepend to all controller routes.
```

README.md
```diff
@@ -431,6 +431,11 @@ App behavior parameters
...
+  
+  - Default: *[String]* `'/'`
+  - Example: `'urlPrefix': '/foo'` would append `foo` too all routes
```
kethinov:
```suggestion
  - Example: `"urlPrefix": "/foo"` would prepend `foo` to all routes.
```

lib/mapRoutes.js
```diff
@@ -177,3 +194,21 @@ module.exports = function (app) {
...
+  // remove whitespace
+  if (typeof urlPrefix === 'string') {
+    urlPrefix = urlPrefix.replace(/\s/g, '')
```
kethinov: why not use `trim()`?

lib/mapRoutes.js
```diff
@@ -177,3 +194,21 @@ module.exports = function (app) {
...
+    // test for valid urlPrefix
+    if (/^[0-9a-zA-Z-_.+!*'(),/$]+$/.test(urlPrefix)) {
+      if (urlPrefix.charAt(0) !== '/') {
+        urlPrefix = [urlPrefix.slice(0, 0), '/', urlPrefix.slice(0)].join('')
```
kethinov: can i assume these changes are not breaking? for instance, if i have an app with a bunch of routes that are already predefined like `/foo/bar` and then i set `urlPrefix` to `/baz`, can i assume that `/foo/bar` becomes `/baz/foo/bar` now, but if i reset `urlPrefix` back to `/` the route still works as originally defined? (`/foo/bar`)

we won't get weird stuff going on like `//foo/bar` right?

README.md
```diff
@@ -431,6 +431,11 @@ App behavior parameters
 
   - Default: *[Number]* `30000` (30 seconds).
 
+- `urlPrefix`: A prefix to append to all routes within the controller files.
+  
+  - Default: *[String]* `'/'`
```
kethinov:
```suggestion
  - Default: *[String]* `"/"`.
```

lib/mapRoutes.js
mulligan121: `.trim()` only removes whitespace from both ends of the string, `.replace()` will allow us to remove the whitespace from the middle of the string as well

lib/mapRoutes.js
mulligan121: This case is handled because anytime `'/'` is the prefix we don't setup a separate router, we use the built in router within `app`. The only issue that someone using roosevelt could potentially run into is the path to the images, stylesheets, or any client-side js files being relative to the route. 
Example:`urlPrefix:  '/baz'` would have the stylesheets located at `/baz/css/*.css` but if the route is `/baz/foo/bar` it might render the view with the path to the stylesheet in `/baz/foo/bar/css/*.css`.
They would be required to prepend the `urlPrefix` by passing it through the global model and to the their views which could get messy. I haven't really figured out a better way to overcome this issue.

README.md
```diff
@@ -444,6 +444,20 @@ App behavior parameters
 
   - Default: *[Number]* `30000` (30 seconds).
 
+- `routers`: An array of objects that define express routers.
```
kethinov:
```suggestion
- `routers`: Here you can define a list of [Express routers](https://expressjs.com/en/guide/routing.html#express-router) which can be used to (among other things) prefix a whole series of routes.
```

README.md
```diff
@@ -444,6 +444,20 @@ App behavior parameters
...
+  - `prefix`: *[String]* The URL path prefix for the router to use.
+  - `controllers`: *[Array]* List of files or directories relative to the controllers directory to use with the defined prefix.
```
kethinov:
```suggestion
  - `controllers`: *[Array]* List of files or directories in `controllersPath` that will be mounted to this route.
```

README.md
```diff
@@ -444,6 +444,20 @@ App behavior parameters
...
+  - Default: *[Boolean]* `false`.
+  - Example:
```
kethinov:
```suggestion
  - Example usage:
```

README.md
```diff
@@ -444,6 +444,20 @@ App behavior parameters
...
+
+    ```json
+    [{
+      "prefix": "/route",
+      "controllers": ["controller.js", "directory"]
+    }]
+    ```
+  **NOTE** When using `routers` roosevelt will modify the `app` argument within controller files to use an express `router`. The app object will be the second argument inside of those controllers.
```
kethinov: This line can go since we're gonna universalize the arguments to always be router and app

README.md
```diff
@@ -444,6 +444,20 @@ App behavior parameters
 
   - Default: *[Number]* `30000` (30 seconds).
 
+- `routers`: Here you can define a list of [Express routers](https://expressjs.com/en/guide/routing.html#express-router) which can be used to (among other things) prefix a whole series of routes.
```
kethinov: Move this section to the bottom of MVC parameters.

kethinov:
```suggestion
- `routers`: Here you can define an array list of [Express routers](https://expressjs.com/en/guide/routing.html#express-router) which can be used to (among other things) prefix a whole series of routes.
```

README.md
mulligan121: This is good now

README.md
mulligan121: This is good now

README.md
```diff
@@ -521,6 +521,34 @@ MVC parameters
   - `serviceUnavailable`: Your [503 Service Unavailable](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#5xx_Server_errors) error page.
     - Default: *[String]* `"503.js"`.
 
+- `routers`: *[Object]* Define routes for express routers and static files
```
kethinov:
```suggestion
- `routers`: *[Object]* List of [Express Routers](https://expressjs.com/en/guide/routing.html#express-router) to create for groups of controllers or statics. If none are defined, Roosevelt will default to creating one single global router with the route prefix `/` that all controllers and all statics will be routed through.
```

README.md
```diff
@@ -521,6 +521,34 @@ MVC parameters
...
+
+      ```json
+      [{
+        "prefix": "/route",
+        "files": ["controller.js", "directory"]
+      }]
+      ```
+
+    - `statics`: *[Array]* Here you can define an array list of objects that represent route prefixes for static files that live inside the `publicFolder`.
+      - `prefix`: *[String]* The URL path prefix for the public directories to use.
```
kethinov:
```suggestion
      - `prefix`: *[String]* The URL path prefix for the public directory to use.
```

README.md
```diff
@@ -521,6 +521,34 @@ MVC parameters
...
+  - Object members:
+    - `controllers`: *[Array]* Here you can define an array list of [Express routers](https://expressjs.com/en/guide/routing.html#express-router) which can be used to (among other things) prefix a whole series of routes.
+      - `prefix`: *[String]* The URL path prefix for the router to use.
+      - `files`: *[Array]* List of files or directories in `controllersPath` that will be mounted to this route.
+      - Default: *[Boolean]* `false`.
+      - Example usage:
+
+      ```json
+      [{
+        "prefix": "/route",
```
kethinov:
```suggestion
        "prefix": "/something",
```

README.md
```diff
@@ -521,6 +521,34 @@ MVC parameters
...
+
+    - `statics`: *[Array]* Here you can define an array list of objects that represent route prefixes for static files that live inside the `publicFolder`.
+      - `prefix`: *[String]* The URL path prefix for the public directories to use.
+      - `files`: *[Array]* List of directories in `publicFolder` that will be mounted to this route.
+      - Default: *[Boolean]* `false`.
+      - Example usage:
+
+      ```json
+      [{
+        "prefix": "/statics",
```
kethinov:
```suggestion
        "prefix": "/something",
```

README.md
```diff
@@ -521,6 +521,34 @@ MVC parameters
...
+      ```json
+      [{
+        "prefix": "/route",
+        "files": ["controller.js", "directory"]
+      }]
+      ```
+
+    - `statics`: *[Array]* Here you can define an array list of objects that represent route prefixes for static files that live inside the `publicFolder`.
+      - `prefix`: *[String]* The URL path prefix for the public directories to use.
+      - `files`: *[Array]* List of directories in `publicFolder` that will be mounted to this route.
```
kethinov:
```suggestion
      - `dirs`: *[Array]* List of directories in `publicFolder` that will be mounted to this route.
```

README.md
```diff
@@ -521,6 +521,34 @@ MVC parameters
...
+      - Default: *[Boolean]* `false`.
+      - Example usage:
+
+      ```json
+      [{
+        "prefix": "/statics",
+        "files": ["css", "images", "js"]
```
kethinov:
```suggestion
        "dirs": ["css", "images", "js"]
```

lib/mapRoutes.js
```diff
@@ -129,50 +131,179 @@ module.exports = function (app) {
...
+              if (requiredController.length > 0 && requiredController.length < 3 && typeof requiredController === 'function') {
+                requiredController(mainRouter, app)
+              }
+            }
+          } catch (e) {
+            if (!ec) {
+              logger.error('🔥', 'The night is dark and full of errors!'.red.bold)
+              ec++
+            }
+            logger.error(`${appName} failed to load controller file: ${controller}. Please make sure it is coded correctly. See documentation at http://github.com/kethinov/roosevelt for examples.\n`.red, e)
```
kethinov:
```suggestion
            logger.error(`${appName} failed to load controller file: ${controller}. Please make sure it is coded correctly. See documentation at http://github.com/rooseveltframework/roosevelt for examples.\n`.red, e)
```

PR #581, opened 2018-11-12
Config Auditor - Skip extra parameters

### LINKED ISSUE #545: Config auditor complains about extra params in multipart object
The root-level `multipart` param is where params for formidable are placed. The config auditor shouldn't care about the stuff in there.

mulligan121:
This makes the config auditor skip the extra parameters check for `multipart`, `bodyParser`, and `logging`. It also skips the missing parameters check for `multipart` and `bodyParser`.

Closes #545

### DIFF
--- a/lib/defaults/config.json
+++ b/lib/defaults/config.json
@@ -7,7 +7,8 @@
     "http": true,
     "appStatus": true,
     "warnings": true,
-    "verbose": false
+    "verbose": false,
+    "_roosevelt-extra-exempt": true
   },
   "minify": true,
   "htmlValidator": {
@@ -25,7 +26,9 @@
     }
   },
   "multipart": {
-    "multiples": true
+    "multiples": true,
+    "_roosevelt-extra-exempt": true,
+    "_roosevelt-missing-exempt": true
   },
   "toobusy": {
     "maxLagPerRequest": 70,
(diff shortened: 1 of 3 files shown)


lib/tools/checkParamObject.js
```diff
@@ -7,9 +7,13 @@ module.exports = function (userParam, defaultParam, auditKey) {
   let defaultProp
   let extraOrMissingkeysBool = false
 
+  // params to skip
+  let skipExtra = ['logging', 'multipart', 'bodyParser']
+  let skipMissing = ['multipart', 'bodyParser']
```
kethinov: let's declare these in config.json

i'm thinking something like this: if we add a new param like `"_roosevelt-exempt": "true"` to things in config.json that can have freeform params, the auditor will simply ignore that param and not audit it

Autre31415: In order to smoothly transition to that you'll have to ensure that the auditor ignores that key, keeping in mind that the auditor and param setter share resources.

lib/defaults/config.json
```diff
@@ -35,7 +38,9 @@
     "urlEncoded": {
       "extended": true
     },
-    "json": {}
+    "json": {},
```
Autre31415: The individual bodyParser params should be exempt, not the bodyParser object itself.

kethinov: ah right, good catch

kethinov: what troy is saying we is we want to require the `json` and `urlEncoded` params, but we don't care what's in them

mulligan121: Ah alright I can do that

Autre31415:
`htmlMinifier.options` should be missing and extra exempt as well.

Issue #305, opened 2018-01-01
Authoring guides for roosevelt-[preprocessors] for CSS and JS


kethinov:
We need `roosevelt-[your CSS preprocessor]` and `roosevelt-[your JS preprocessor]` module authoring guides. This shouldn't be too difficult to write documentation for as the API is pretty thin and we have some reference implementations. But even so it should get documented.

PR #605, opened 2018-12-13
User-defined CSS & JS preprocessors

### LINKED ISSUE #599: Option to supply custom CSS/JS preprocessor middleware on the fly at runtime
In `package.json`:

```
    "css": {
      "compiler": {
        "nodeModule": "custom",
     [...]
    "js": {
      "compiler": {
        "nodeModule": "custom",
```

If a custom middleware is declared, then you need to supply a custom middleware code at start time via the roosevelt constructor that looks like this:

```js
let app = require('roosevelt')({
  cssCompiler: (app) => {
    return {
      versionCode: (app) => {
        // functionality
      },
      parse: (app, fileName) => {
        // functionality
      }
    }
  },
  jsCompiler: (app) => {
    return {
      parse: (app, fileName) => {
        // functionality
      }
    }
  }
})
```

Since authoring the custom middleware requires documentation, this should probably be paired with https://github.com/rooseveltframework/roosevelt/issues/305

dishad:
Closes #599 and #305

### DIFF
--- a/roosevelt.js
+++ b/roosevelt.js
@@ -35,6 +35,8 @@ module.exports = function (params) {
   let clusterKilled = 0
   let checkConnectionsTimeout
   let shutdownType
+  let cssProcessor
+  let jsProcessor
 
   // expose initial vars
   app.set('express', express)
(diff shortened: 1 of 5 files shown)


roosevelt.js
```diff
@@ -169,6 +171,16 @@ module.exports = function (params) {
     params.onServerInit(app)
   }
 
+  // use user-defined cssCompiler
+  if (params.cssCompiler && typeof params.cssCompiler === 'function') {
+    cssProcessor = params.cssCompiler(app)
```
kethinov: this work should be done in sourceParams instead of doing it here and then passing it as an argument to the compiler modules

the individual compiler modules can get the compiler out of params later

PR #609, opened 2018-12-20
Reports files without Express routes in controllers

### LINKED ISSUE #441: Consider scanning for the existence of Express routes in file before loading it as a controller
An extension of #122

dishad:
Closes #441

### DIFF
--- a/lib/checkRoutes.js
+++ b/lib/checkRoutes.js
@@ -1,0 +1,41 @@
+require('colors')
+const fs = require('fs-extra')
+const path = require('path')
+const logger = require('./tools/logger')()
+
+// Utility module for checking express routes in controllersPath
+module.exports = function (app, callback) {
+  const controllersPath = app.get('controllersPath')
+  let badControllers = []
+  let controllerFiles
+
+  // If there is no controllers directory, just continue
+  try {
+    fs.accessSync(controllersPath)
+  } catch (e) {
+    callback()
+  }
+
+  // Get all controller files
+  controllerFiles = fs.readdirSync(controllersPath)
+
+  // Loop through files and their content
+  for (let controller of controllerFiles) {
+    let file = path.join(controllersPath, controller)
+    let content = fs.readFileSync(file, 'utf8')
+
+    // If no express routes are found (error files may not have an app.route())
+    if (!content.match(/\.(route|status|send)*?\(([^)]*)\)/)) {
+      badControllers.push(controller)
+    }
+  }
+
+  // Report bad controller files
+  if (badControllers.length > 0) {
+    badControllers[0] = '       - ' + badControllers[0]
+    badControllers = badControllers.join('\n       - ')
+    logger.log('⚠️', `There are files without Express routes located in your controllers directory.\n${badControllers}`.yellow)
+  }
+
+  callback()
+}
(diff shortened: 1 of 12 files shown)


lib/checkRoutes.js
```diff
@@ -0,0 +1,41 @@
...
+    if (!content.match(/\.(route|status|send)*?\(([^)]*)\)/)) {
+      badControllers.push(controller)
+    }
+  }
+
+  // Report bad controller files
+  if (badControllers.length > 0) {
+    badControllers[0] = '       - ' + badControllers[0]
+    badControllers = badControllers.join('\n       - ')
+    logger.log('⚠️', `There are files without Express routes located in your controllers directory.\n${badControllers}`.yellow)
```
kethinov:
```suggestion
    logger.warn(`There are files without Express routes located in your controllers directory.\n${badControllers}`.yellow)
```

dishad: Current PR should reflect these changes

kethinov:
Looks like a coverage drop.

dishad:
Should now achieve 100% coverage

lib/checkRoutes.js
```diff
@@ -0,0 +1,41 @@
...
+    // Get all controller files
+    controllerFiles = fs.readdirSync(controllersPath)
+
+    // Loop through files and their content
+    for (let controller of controllerFiles) {
+      let file = path.join(controllersPath, controller)
+      let content = fs.readFileSync(file, 'utf8')
+
+      // If no express routes are found (error files may not have an app.route())
+      if (!content.match(/\.(route|status|send)*?\(([^)]*)\)/)) {
```
kethinov: i think there might be some false negatives here

take a look at the [various ways to write express routes](https://expressjs.com/en/guide/routing.html)

wouldn't this test fail something like this?

```javascript
  app.get('/stuff', (req, res) => {
    res.render('something', {})
  })
```

lib/checkRoutes.js
```diff
@@ -0,0 +1,41 @@
...
+  const controllersPath = app.get('controllersPath')
+  let badControllers = []
+  let controllerFiles
+
+  // Make sure controllersPath is a directory
+  try {
+    fs.accessSync(controllersPath)
+
+    // Get all controller files
+    controllerFiles = fs.readdirSync(controllersPath)
```
kethinov: needs to be recursive

lib/checkRoutes.js
```diff
@@ -0,0 +1,41 @@
...
+
+    // Loop through files and their content
+    for (let controller of controllerFiles) {
+      let file = path.join(controllersPath, controller)
+      let content = fs.readFileSync(file, 'utf8')
+
+      // If no express routes are found (error files may not have an app.route())
+      if (!content.match(/\.(render|status|send)*?\(([^)]*)\)/)) {
```
kethinov: this doesn't handle scenarios like this:

```js
// 404.js
const notFound = require('controllers/notFound')

module.exports = app => {
  app.all('*', (req, res) => {
    notFound(app, req, res)
  })
}
```

kethinov:
While testing this I've seen some valid use cases for writing controllers without routes in them. I'm gonna close this PR for now.

PR #616, opened 2018-12-28
Added reload auto browser refresher

### LINKED ISSUE #32: Browser auto-reload dev mode feature

alallier:
Added [reload ](https://www.npmjs.com/package/reload) as a dev mode feature to Roosevelt. This automatically refreshes the web pages on server restart.

(Closes #32)


alallier:
I haven't merged on the reload side yet, because I want to make sure we are set on this side first

roosevelt.js
```diff
@@ -362,9 +395,23 @@ module.exports = function (params) {
...
+          })
+        }
       }
       if (httpsParams.enable) {
         serverPush(httpsServer, httpsParams.port, 'HTTPS')
+        if (httpsReload && params.reload.enabled === true) {
+          httpsReload.startWebSocketServer().then(function () {
+            logger.log('🎧', `Reload HTTPS server is listening on port: ${params.reload.port}`.bold)
+          }).catch(function (err) {
+            logger.error(('Reload was unable to start due to an error' + err.toString()).red)
```
alallier: I wanted to do `logger.error(('Reload was unable to start due to an error', err).red)` with string interpolation but that didn't seem to work

alallier: Also I didn't add an emoji, you can tell me what one you want me to do.

alallier:
This needs unit tests

lib/sourceParams.js
```diff
@@ -72,6 +72,7 @@ module.exports = function (app) {
     shutdownTimeout: checkParams(params.shutdownTimeout, pkg.rooseveltConfig.shutdownTimeout, defaultConfig.shutdownTimeout),
     cores: checkParams(params.cores, pkg.rooseveltConfig.cores, defaultConfig.cores),
     cleanTimer: checkParams(params.cleanTimer, pkg.rooseveltConfig.cleanTimer, defaultConfig.cleanTimer),
+    reload: checkParams(params.reload, pkg.rooseveltConfig.reload, defaultConfig.reload),
```
alallier: You probably want to make this name tool agnostic. Feel free to tell me what you want to change it to and I can make the change.

package.json
```diff
@@ -39,6 +39,7 @@
     "parent-require": "~1.0.0",
     "prismjs": "~1.15.0",
     "ps-node": "~0.1.6",
+    "reload": "git+https://github.com/alallier/reload.git#3.0.0",
```
alallier: This needs to change, only here so that I can make sure we are set on integration before merging on the reload side

lib/sourceParams.js
kethinov: Let's go with `frontendReload`. This distinguishes it from a backend reloader like `nodemon`, and we can draw that distinction in the README.

roosevelt.js
kethinov: IIRC the logger adds the default emoji for errors, yes?

lannonbr: Not sure if it does in the logger in roosevelt right now does, but the extracted roosevelt-logger module does

Autre31415: So something about combining strings, error objects, and colors in a single log messes with the expansion of the object, so in general across the codebase we just drop in an additional `logger.error(err)` so that the user can see the full stack trace.

lib/sourceParams.js
alallier: Updated

roosevelt.js
alallier: Thanks @Autre31415. Will leave final logging to @kethinov

alallier:
Another point we need to think about here is reload won't work without having the [reload required script tag](https://github.com/alallier/reload#express-example) present in the HTML.

We can make the change in the generator but previous applications and manual removal will make this feature stop working

There is server side injection could get tricky... On the reload end we have considered supporting server side injection but have not implemented it

package.json
kethinov: still waiting on this

but is everything else done?

alallier: Was waiting for final word on this end before merging reload.

Still need unit tests as well.

kethinov:
You could do server-side objection within roosevelt using the same infrastructure that is used to inject validator output in dev mode. Use the Express render/send override technique to force in the script tag when `frontendReload` is enabled. You can scan the markup for `</body>` and toss the script tag right before that string. If you can't find a `</body>` then dump it at the end of the string and hope for the best maybe?

alallier:
We could give that a try and go from there?

package.json
alallier: This branch is published now. I will make the update

lib/defaults/config.json
```diff
@@ -47,6 +47,12 @@
       "_roosevelt-extra-exempt": true
     }
   },
+  "frontendReload": {
+    "enabled": true,
```
Autre31415: Change this to `enable`

roosevelt.js
```diff
@@ -8,6 +8,7 @@ const path = require('path')
 const os = require('os')
 const fs = require('fs')
 const fsr = require('./lib/tools/fsr')()
+const reload = require('reload')
```
Autre31415: We want to ensure that all dev dependencies are isolated behind dev mode flags in such a way that one could delete all dev dependencies and still operate a roosevelt app without errors. To that effect all references to reload and its functionality should ideally be isolated into their own file if possible, or at the very least buried under a development mode check.

lib/defaults/config.json
kethinov:
```suggestion
    "enable": true,
```

package.json
Autre31415: I don't believe pointing this to git is necessary anymore

roosevelt.js
dishad: Now hides all calls to dev dependencies behind dev mode flags

PR #617, opened 2019-01-03
Add OpenJDK support to HTML Validator

dwahrens:
HTML Validator would not start when using OpenJDK. Change fixes so Roosevelt can now run with OpenJDK.


lannonbr:
LGTM. @kethinov, can you take a quick check to verify as well.

PR #650, opened 2019-02-26
Add Multiple exceptions to HTML validator 

### LINKED ISSUE #607: Support multiple exceptions in htmlValidator
Validator should not trigger on non `res.render`s. In the case of `res.send` (or others) the validator should not trigger because it will always be invalid.

I created a [sample repository](https://github.com/alallier/validator-test) to demo this.

Start the app and hit the `/test` route in _both_ **`dev`** and **`prod`** mode

### LINKED ISSUE #651: TypeError if the model object is omitted when using res.render()
According to the [Express documentation](https://expressjs.com/en/api.html#res.render), `res.render()` has an optional second argument called `locals` (roosevelt documents it as model).

If  we omit the second argument, roosevelt will throw an error. `❌  TypeError: Cannot read property '_disableValidator' of undefined`

Steps to reproduce:
1. Use `generator-roosevelt` to create a new roosevelt application
2. Remove the argument `model` from the homepage controllers `res.render`
3. Spin up the app
4. http://localhost:43711/ in the browser
5. There will be a 500 response with the error in the console

I tested this on versions 0.12.0 & 0.13.0

lyomr:
- Modified htmlValidator to accept multiple exceptions
- Closes #607 
- Added missing flags from sources
- updated CHANGELOG
- Fix TypeError (closes #651)


kethinov:
You should fix this as part of this PR: https://github.com/rooseveltframework/roosevelt/issues/651

lib/htmlValidator.js
```diff
@@ -137,9 +137,13 @@ module.exports = function (app, callback) {
...
     let render = app.response.render
 
+    if (Array.isArray(ValidatorExceptions) && ValidatorExceptions.length >= 1) {
+      console.log('my exceptions', ValidatorExceptions[1].requestHeader)
+    } // check if there are multiple exceptions
+
     app.response.render = function (view, model, callback) {
-      if (model[modelException]) {
-        this.req.headers[headerException] = true
+      if (model[ValidatorExceptions.modelValue]) {
```
mulligandev: I think this is the line throwing the type error in issue #651

kethinov:
closing in favor of https://github.com/rooseveltframework/roosevelt/pull/672

PR #672, opened 2019-03-28
Add Multiple exceptions to HTML validator

### LINKED ISSUE #607: Support multiple exceptions in htmlValidator

### LINKED ISSUE #651: TypeError if the model object is omitted when using res.render()

lyomr:
- Modified htmlValidator to accept multiple exceptions (Closes #607)
- Added missing flags from sources and appropriate tests
- updated CHANGELOG
- Fixes TypeError (Closes #651)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,8 @@
 
 ## Next version
 
-- Enhancement to `res.redirect()` for custom routers to prepend the prefix to redirects that are relative to the hostname
+- Enhancement to `res.redirect()` for custom routers to prepend the prefix to redirects that are relative to the hostname.
+- Add option to allow multiple exceptions to html validator.
 
 ## 0.13.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,9 @@
 ## Next version
 
 - Enhancement to `res.redirect()` for custom routers to prepend the prefix to redirects that are relative to the hostname.
-- Add option to allow multiple exceptions to html validator.
+- HTML validator `exceptions` param will now accept an array of values instead of just a single string value.
+- Various dependencies bumped.
+- CI improvements. 
 
 ## 0.13.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Next version
 
-- Enhancement to `res.redirect()` for custom routers to prepend the prefix to redirects that are relative to the hostname.
+- If custom routers are being used, the `res.redirect()` method will now prepend the prefix to redirects that are relative to the hostname. To override this setting pass `true` as the last argument.
 - HTML validator `exceptions` param will now accept an array of values instead of just a single string value.
 - Various dependencies bumped.
 - CI improvements. 
--- a/lib/htmlValidator.js
+++ b/lib/htmlValidator.js
@@ -210,6 +210,7 @@ module.exports = function (app, callback) {
         headerException.forEach(function (exception) {
           // check the request header
           if (req.headers[exception.toLowerCase()]) {
+            res.set(exception.toLowerCase(), true)
             found = true
           }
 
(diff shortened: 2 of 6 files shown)


lib/htmlValidator.js
```diff
@@ -138,9 +138,39 @@ module.exports = function (app, callback) {
...
-        this.req.headers[headerException] = true
+      if (!model) {
+        model = {}
       }
+
+      let requestHeaders = this.req.headers
+      if (modelException) {
+        if (typeof modelException === 'string') {
+          if (model[modelException]) {
+            applyHeaderExceptions()
```
Autre31415: Simple refactor that could possibly make the code coverage aspect easier:

Rather than testing exceptions for both `string` and `array` config examples, coerce your `string` scenario into an array, then you only have to run the match logic and execute the header exception function in one place.

You could do this by caching your exceptions in a new variable that is initialized as an empty array, then either pushing the string into it if the param is a string, or just setting it to whatever the array is in the case of it being configured as an array, then run your model/header exception loop as normal.

CHANGELOG.md
```diff
@@ -2,7 +2,8 @@
 
 ## Next version
 
-- Enhancement to `res.redirect()` for custom routers to prepend the prefix to redirects that are relative to the hostname
+- Enhancement to `res.redirect()` for custom routers to prepend the prefix to redirects that are relative to the hostname.
```
kethinov:
```suggestion
- If custom routers are being used, the `res.redirect()` method will now prepend the prefix to redirects that are relative to the hostname. To override this setting pass `true` as the last argument.
```

CHANGELOG.md
```diff
@@ -2,7 +2,8 @@
...
+- Add option to allow multiple exceptions to html validator.
```
kethinov:
```suggestion
- HTML validator `exceptions` param will now accept an array of values instead of just a single string value.
- Various dependencies bumped.
- CI improvements. 
```

lib/htmlValidator.js
```diff
@@ -186,11 +198,35 @@ module.exports = function (app, callback) {
         return validationMessage
       }
 
-      if (req.headers[headerException]) {
-        res.set(headerException, true)
+      function validatorExceptions () {
+        let found = false
+
+        if (foundModelException) {
+          found = true
```
kethinov: why bother with the rest of the function? just return true here

Autre31415: Good point. Get rid of the found variable and just replace every instance of `found = true` with `return true` and then replace drop a `return false` under the exception checks.

lib/htmlValidator.js
```diff
@@ -186,11 +198,28 @@ module.exports = function (app, callback) {
...
+          return true
+        }
+
+        headerException = [].concat(headerException)
+
+        headerException.forEach(function (exception) {
+          // check the request header
+          if (req.headers[exception.toLowerCase()]) {
+            res.set(exception.toLowerCase(), true)
+            return true
```
kethinov: the reason the found variable was used is this won't work

you're returning inside the forEach callbac, not the parent function

PR #709, opened 2019-05-17
Move all html validator dependencies into devdeps

### LINKED ISSUE #708: Move the HTML validator functionality to its own module
Move the HTML validator functionality to its own module.

Right now the validator is set to only run in development mode which makes plenty of sense. But it would be nice to create a new module to house this code, so that when we do a `npm install --production` the code for the validator will not be present on the system.

This will help remove code delivered in the production environment that we don't care about, and limit possible security concerns of having code that spawns and kills a process on the system.

Autre31415:
By ensuring the htmlValidator.js file only executes in dev mode we are able to move the following dependencies into devdeps:
- fkill
- html-validator
- prismjs
- ps-node
- tamper
- tmp
- vnu-jar

Closes #708


lib/htmlValidator.js
```diff
@@ -38,9 +38,6 @@ module.exports = function (app, callback) {
     validatorOptions.port = params.htmlValidator.port
```
alallier: Will this trump the node ENV?

alallier: I know source configs trumps whatever the validator is set to with the Node ENV but we aren't checking the Node ENV here

alallier: [https://github.com/rooseveltframework/roosevelt/blob/master/lib/sourceParams.js#L125-L149](HTML Validtor source configs) is trumped finally at the end of source configs by [this](https://github.com/rooseveltframework/roosevelt/blob/master/lib/sourceParams.js#L155-L157).

alallier: Yeah sorry I was reading this wrong, nevermind. Looks fine

PR #724, opened 2019-05-31
Added scripts to install or remove devdeps

### LINKED ISSUE #720: Script to remove dependencies only needed in dev mode for production builds
The devDependencies approach attempted in https://github.com/rooseveltframework/roosevelt/issues/708 turned out to be problematic, so only half a solution was implemented. If any dependencies only needed for dev mode are missing, the app will not crash in production mode. That is progress, but we still need a convenient way to actually remove dev-only dependencies from apps without having to manually identify and remove them. After discussing this with @alallier, we propose the following:

1. Make a whitelist of (direct) dependencies that are only needed in dev mode.
2. Explore feasibility of using [npm uninstall](https://docs.npmjs.com/cli/uninstall) to remove them at the app level.
3. For npm uninstall to be considered feasible, the removal should remove both the direct dependency and its subdependencies.
4. Crucially, it should *not* remove subdependencies that are shared with other dependencies that are *not* being removed.

Simplified example:

SimpliRoosevelt dependency graph:

1. Foo: depends on Bar.
2. Baz: depends on Bar and Zog.

If we use npm uninstall on Baz only, the behavior we want is for it to remove Baz and Zog because Zog is a dependency of Baz only. But it should *not* remove Bar because Bar is a dependency of Foo, which we are not removing.

If npm uninstall functions this way (or a similar npm command functions this way), then it would be pretty trivial for us to maintain a script that can remove deps that are not needed in production mode.

reecer62:
Closes #720

Moved the following deps back into devdeps:
1. fkill
2. html-validator
3. prismjs
4. ps-node
5. tamper
6. tmp
7. vnu-jar

- Implemented "npm run dev-install" which will install dev deps of roosevelt
- Implemented "npm run dev-prune" which will remove dev deps of roosevelt
- Added to postinstall script, now dev deps will be automatically installed unless an environment variable "ROOSEVELT_DEPLOYMENT" is set

Here are two tests of running 

> npm i

 in a newly generated app, one with a deployment variable, and one without.
720-skipdevdeps
720-installdevdeps


lib/scripts/getDevDeps.js
```diff
@@ -0,0 +1,16 @@
+const execSync = require('child_process').execSync
+
+/**
+ * check for deployment environment variable ROOSEVELT_DEPLOYMENT
+ * if it exists don't install roosevelt devdeps
+ * otherwise, install devdeps
+ * @function installDevDeps
+ */
+;(function installDevDeps () {
```
Autre31415: This doesn't need to be a self-executing function. In fact it doesn't need to live inside a function at all since it's encapsulated in this script file anyway.

lib/scripts/getDevDeps.js
```diff
@@ -0,0 +1,16 @@
...
+  if (process.env.ROOSEVELT_DEPLOYMENT) {
+    console.log('Skipping roosevelt dev dependencies.')
```
Autre31415: Use the logger feature for these logs. [Example](https://github.com/rooseveltframework/roosevelt/blob/master/lib/scripts/configAuditor.js#L56)

kethinov: make sure to inject suitable emojis too

PR #759, opened 2019-07-05
WIP: Make tests more stable

reecer62:
Removed instances where a kill signal was being sent to the app when the app server isn't running. If certain race conditions met, the kill signal would be sent after test closed resulting in an error.
Todo:
- [x] do some more research into how conditions are created
- [x] sometimes test app not killed in validator tests, fix


reecer62:
The code for shutting down the server is not hit occasionally, leaving the port in use even though the kill signal was sent. Also, it seems like the only consistently failing tests that are left are the ones testing the `cores` flag in `rooseveltTests.js`. The code for core usage is also in the same area of code as the shutdown code. I will keep investigating to diagnose what exactly is causing the app to not fully shutdown.

PR #763, opened 2019-07-09
Fixed getDevDeps script so the postinstall script will work correctly

### LINKED ISSUE #762: Possible to lose your dev dependencies
Here are the reproduction steps:

1. Create sample app
1. Install dependencies (without any ENVs set) `npm i`
1. Start app, everything is running fine
1. Install dependencies again (without any ENVs set) `npm i`
1. Start app, broken due to missing dev dependencies

(My hunch) This happens because Roosevelt is already installed when running the npm install again the second time so it's post install script is not ran. None of Roosevelt's dev dependencies are found in the package-lock because they are installed via a post install script the first time. Since the dev dependencies are not in the package-lock the _second_ `npm install` deletes them, leaving you in a non-intuitive broken state

reecer62:
(Closes #762)
Checks in Roosevelt's node_modules folder for one of its dev dependencies before installing devDeps, as long as devDeps aren't already installed it will install them as expected

Todo:
- [x] add to CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Next version
 
+- Fixed a bug where postinstall script would remove devDeps on an additional run of `npm i`
 - Postinstall script checks if Roosevelt itself is a devDependency of the app before installing its own devDeps (no longer installs Roosevelt devDeps when running `npm i` in Roosevelt sub modules)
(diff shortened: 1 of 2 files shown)


alallier:
Thanks for the quick fix on this guys, truly appreciated!

PR #770, opened 2019-07-15
Reworked config auditor to check config params on a case by case basis

### LINKED ISSUE #719: Make config auditor smarter
The config auditor is great but it needs to be a little smarter.

Specifically when it is checking for extra params it has to check against a value that represents the largest possible option.

For example, the HTTPS config used to look like this. (This is still what the yeoman generator will give you for example)

```json
"https": {
  "enable": false,
  "httpsOnly": false,
  "httpsPort": 43733,
  "pfx": false,
  "keyPath": "null",
  "passphrase": "null",
  "ca": "null",
  "cafile": true,
  "requestCert": false,
  "rejectUnauthorized": false
}
```

But some parts of this configuration are no longer valid. This doesn't error because the configuration auditor checks the key https to see if it is [missing](https://github.com/rooseveltframework/roosevelt/blob/master/lib/scripts/configAuditor.js#L59-L73) (which it isn't) and it checks to see if it is [_extra_(/undefined)](https://github.com/rooseveltframework/roosevelt/blob/master/lib/scripts/configAuditor.js#L76-L81) (which it also isn't because HTTPS is defined).

When doing the config audit we should check against a value that contains all of the possible key/value pairs to flags on items that are outdated.

reecer62:
Closes #719 

Todo:

- [x] CHANGELOG.md
- [x] wait for merge of [source-configs](https://github.com/rooseveltframework/roosevelt/pull/775) internal sourcing of params
- [x] delete checkParamObject.js (when ^ is merged, this file will no longer be used anywhere)
- [x] tests
- [x] type checker
- [x] remove default `config.json` file
- [x] update tests
- [x] update documentation with sample config

Example test config and the output from running the reworked auditor:
```
"rooseveltConfig": {
    "extraParam": true,
    "enableCLIFlags": true,
    "generateFolderStructure": true,
    "localhostOnly": true,
    "routers": false,
    "cleanTimer": 604800000,
    "logging": {
      "methods": {
        "http": true,
        "info": true,
        "warn": true,
        "error": true,
        "verbose": false
      }
    },
    .
    .
    . (rest of the default generated config here)
  },
  "scripts": {
    "development": "nodemon app.js --development-mode",
    "dev": "nodemon app.js --development-mode",
    "d": "nodemon app.js --development-mode",
    "dev-install": "roosevelt cd ./node_modules/roosevelt && npm install --only=dev",
    "di": "cd ./node_modules/roosevelt && npm install --only=dev",
    "dev-prune": "cd ./node_modules/roosevelt && npm prune --production",
    "dp": "cd ./node_modules/roosevelt && npm prune --production",
    "kill-validator": "node ./node_modules/roosevelt/lib/scripts/killValidator.js",
    "kv": "node ./node_modules/roosevelt/lib/scripts/killValidator.js",
    "clean": "node ./node_modules/roosevelt/lib/scripts/appCleanup.js",
    "config-audit": "node ./node_modules/roosevelt/lib/scripts/configAuditor.js",
    "standard": "standard",
    "stylelint": "./node_modules/.bin/stylelint \"statics/css/**/*.less\" --syntax less",
    "lint": "standard && ./node_modules/.bin/stylelint \"statics/css/**/*.less\" --syntax less",
    "test": "echo \"TODO: tests\""
  },
```
Output:
```
Starting rooseveltConfig audit...
 Missing param "port"!
 Extra param "extraParam" found at rooseveltConfig.extraParam, this can be removed.
 Detected outdated script "dev-install". Update contents to "cd ./node_modules/roosevelt && npm install --only=dev" to restore functionality.
 Missing production run script(s).
Issues have been detected in rooseveltConfig, please consult https://github.com/rooseveltframework/roosevelt#configure-your-app-with-parameters for details on each param.
Or see https://github.com/rooseveltframework/roosevelt/blob/master/lib/defaults/config.json for the latest sample rooseveltConfig.
```


reecer62:
I used the Roosevelt docs on app behavior parameters but the individual cases for each param should be reviewed to confirm whether or not each param should be ignored/checked if missing from user config and whether or not to ignore extra params in the param (if an object).

Some examples:

-  the `methods` object can hold custom log types so the auditor should ignore extra params in that object because there can be user-defined params in it.
-  the `port` param should be checked if it is missing from the user config and should warn the user
-  the `options` object in the `htmlMinifier` object is not necessary, so there is not a check for if the param is missing
-  the `bodyParser` object should be checked for extra params because its params, `urlEncoded` and `json`, are the only params consumed within that object by Roosevelt

reecer62:
Once [PR#775](https://github.com/rooseveltframework/roosevelt/pull/775) is merged, this PR will result in the sourcing of params and the auditing of params to be completely separated. Previously, both used the checkParamObject.js file which ended up in both loading params. Now that this is not the case, the update to sourcing params with source-configs is necessary to pass the "params" tests (which test loading params from app, package.json, and defaults)

reecer62:
Is ready for review

lib/scripts/configAuditor.js
```diff
@@ -47,78 +44,555 @@ function configAudit (appDir) {
...
+              break
+            case 'exceptions':
+              const exceptionsParam = validatorParam[validatorKey] || {}
+              const defaultExceptionsParam = defaultValidatorParam[validatorKey]
+              const exceptionsKeys = Object.keys(defaultExceptionsParam)
+              exceptionsKeys.forEach((exceptionKey) => {
+                keyStack.push(exceptionKey)
+                switch (exceptionKey) {
+                  case 'requestHeader':
+                    delete exceptionsParam[exceptionKey]
```
lannonbr: is it correct that a case like this where you just delete the key means that the key is optional?

lib/scripts/configAuditor.js
```diff
@@ -47,78 +44,555 @@ function configAudit (appDir) {
...
+   * Recursive function that checks for extra params and lets the user know where to find them
+   *
+   * @param userConfigParam - extra params in the userConfig
+   * @param defaultConfigParam - object in the defaultConfig at the key specified in the userConfigParam
+   * @param jsonPath - string containing the path to the extra param
+   */
+  function checkExtraParams (userConfigParam, defaultConfigParam, jsonPath) {
+    const params = Object.entries(userConfigParam)
+    params.map(([key, val]) => {
+      if (val instanceof Object && !(Array.isArray(val)) && val !== null && defaultConfigParam[key] !== undefined) {
```
lannonbr: May be possibly good to leave a comment for this conditional that this is when we have a further layer of params we want to examine.

lib/scripts/configAuditor.js
```diff
@@ -47,78 +44,555 @@ function configAudit (appDir) {
...
+          errors = true
+        }
+        break
     }
   })
 
   // output warning for missing start script
   if (!hasProdScript) {
-    logger.warn('Missing production run script(s).'.bold)
+    logger.warn(' Missing production run script(s).'.bold)
```
lannonbr: Extra space here on accident or to properly align the content with the emojis on the left? Is this properly formatted when emojis are disabled?

reecer62: The space was to align the text with other `logger.warn` and `logger.error` statements. All of them have a space before the text

lib/scripts/configAuditor.js
reecer62: I will add a comment

lib/scripts/configAuditor.js
reecer62: The purpose of delete on a param is to prevent it from being considered an extra param. This is because whatever params are left in the `userConfig` are considered extra params

kethinov: This creates the following regression:

1. Create a fresh sample app
2. Set the roosevelt dependency to your branch
3. `npm i`
4. `npm run dev`

App will crash because you deleted a bunch of keys from the config...

You can fix this by copying the object instead of passing by reference.

reecer62: Updated

reecer62:
The list of types that params should be checked for should be reviewed and the list of types passed in for every param should be reviewed too

PR #782, opened 2019-07-29
Check non-devDeps only when `--only=prod` is set

### LINKED ISSUE #771: Check dependencies will give false positive when using `npm install --only=prod`
If you don't have the dev dependencies installed (probably due to having them skipped for a deployment) the [check dependencies module](https://github.com/rooseveltframework/roosevelt/blob/11820b17dc6cc6548a37871717e99be6ce7a91dd/roosevelt.js#L52-L60) that Roosevelt uses will give a false warning

Check dependencies has a [scope option](https://www.npmjs.com/package/check-dependencies#scopelist), during this mode we need to change the scope. Not sure exactly how we will know when we are in this mode, but we could use best guess and tell the user it's best guess given not a single dev dependency is installed. "No development dependencies are installed, this is okay if in a deployment. If you want to install the development dependencies run the script" or something like that.

reecer62:
Closes #771 
Todo:

- [x] Update CHANGELOG.md
- [x] Tests


reecer62:
> 
> 
> You also need to handle the check in `app.js`

Are you referring to the `app.js` file in the generated app?

alallier:
> > You also need to handle the check in `app.js`
> 
> Are you referring to the `app.js` file in the generated Roosevelt app?

@reecer62 sorry I meant to say roosevelt.js. We need a modification on the application running [`check dependencies`](https://github.com/rooseveltframework/roosevelt/blob/037fd359f2cd95219b62bfec459b153e8d422387/roosevelt.js#L52-L60)

alallier:
@reecer62 had another thought, this needs to support all ways someone can set prod/dev mode. Setting via ENV variables will need to be accounted for as well

reecer62:
> 
> 
> @reecer62 had another thought, this needs to support all ways someone can set prod/dev mode. Setting via ENV variables will need to be accounted for as well

I am having trouble understanding why/what cases could arise that would require these additional checks

CHANGELOG.md
```diff
@@ -1,4 +1,5 @@
 # Roosevelt Changelog
+- Updated Roosevelt to check non-dev dependencies when Roosevelt app is ran
```
kethinov: this should be moved into 0.15.0

reecer62: Done

PR #825, opened 2019-11-18
warn when app directory has been moved

lyomr:


lib/tools/fsr.js
```diff
@@ -24,6 +26,7 @@ class FSR {
       try {
         fs.symlinkSync(...arguments)
       } catch (e) {
+        logger.error('Your Roosevelt App has been moved. You may want to run \'npm run clean\' to remake the symlinks.\n')
```
kethinov:
```suggestion
        logger.error('It appears your Roosevelt app has been moved. You may want to run `npm run clean` to remove the broken symlinks.\n')
```

kethinov: Also make sure to break out this catch block to sniff for the specific error code that triggers when the app is moved. We don't want to show this error message every time the try/catch fails.

PR #875, opened 2020-03-26
Replace routers param with a much simpler routePrefix param

Autre31415:
- Replace `routers` param with a simplified `routePrefix` param which defines a subfolder to mount the app to.
- Add `router` express variable that opens up access to internally used router instance.
- Add `routePrefix` express variable that exposes currently used routePrefix or an empty string if not used.


README.md
```diff
@@ -606,50 +606,13 @@ Resolves to:
...
-      ```
-
-    - `public`: *[Array]* List of [Express Routers](https://expressjs.com/en/guide/routing.html#express-router) which can be used to (among other things) prefix a whole series of static files into a series of public routes.
-
-      - `prefix`: *[String]* The URL path prefix for the public directory to use.
-
-      - `dirs`: *[Array]* List of directories in `publicFolder` that will be mounted to this route.
-
-      - Default: *[Boolean]* `false`.
+  - Example: When set to `"foo"` a route bound to `/` will be responded to at `/foo/`.
```
kethinov:
```suggestion
  - Example: When set to `"foo"` a route bound to `/` will be instead be bound to `/foo/`.
```

README.md
```diff
@@ -606,50 +606,13 @@ Resolves to:
...
 
-      - Example usage:
+  - Note: This prefix is exposed via the `routePrefix` express variable which can be handy for resolving the absolute paths to statics programmatically.
```
kethinov:
```suggestion
  - Note: This prefix is exposed via the `routePrefix` Express variable which should be used for resolving the absolute paths to statics programmatically.
    - TODO: show example of this.
```

CHANGELOG.md
```diff
@@ -7,6 +7,9 @@
 - Added ability to configure Roosevelt via a rooseveltConfig.json config file placed in app root.
 - Renamed `multipart` param to `formidable`.
 - Updated internal usage of formidable API.
+- Replace `routers` param with a simplified `routePrefix` param which defines a subfolder to mount the app to.
```
kethinov:
```suggestion
- Replaced `routers` param with a simplified `routePrefix` param which defines a subfolder to mount the app to.
```

CHANGELOG.md
```diff
@@ -7,6 +7,9 @@
...
+- Add `router` express variable that opens up access to internally used router instance.
```
kethinov:
```suggestion
- Added `router` Express variable that opens up access to internally used router instance.
```

CHANGELOG.md
```diff
@@ -7,6 +7,9 @@
...
+- Add `routePrefix` express variable that exposes currently used routePrefix or an empty string if not used.
```
kethinov:
```suggestion
- Added `routePrefix` Express variable that exposes currently used routePrefix or an empty string if not used.
```

PR #882, opened 2020-04-08
Remove TODO from readme

### LINKED ISSUE #876: Resolve the TODO in the README

Autre31415:
Closes #876

### DIFF
--- a/README.md
+++ b/README.md
@@ -611,7 +611,8 @@ Resolves to:
   - Example: When set to `"foo"` a route bound to `/` will be instead be bound to `/foo/`.
 
   - Note: This prefix is exposed via the `routePrefix` Express variable which should be used for resolving the absolute paths to statics programmatically.
-    - TODO: show example of this.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -611,7 +611,8 @@ Resolves to:
   - Example: When set to `"foo"` a route bound to `/` will be instead be bound to `/foo/`.
 
   - Note: This prefix is exposed via the `routePrefix` Express variable which should be used for resolving the absolute paths to statics programmatically.
-    - TODO: show example of this.
+
+    - Example: An image located at `images/teddy.jpg` can be resolved in a prefix agnostic way via `` `${app.get('routePrefix')/images/teddy.jpg}` ``.
```
kethinov, line 615: I *think* this suggestion is right?

```suggestion
    - Example: An image located at `/images/teddy.jpg` can be resolved in a prefix agnostic way via `` `${app.get('routePrefix')/images/teddy.jpg}` ``.
```

Also is it right to assume that if we're passing the prefix to a template, that would need to be passed down in the model and end up looking like something like ``?

Autre31415: Yeah for a template you'd need to handle it that way.

PR #892, opened 2020-04-17
Properly fix CSS bug

alallier:
I don't know what happened here, I'm really perplexed how I saw this working in my original fix #885. Maybe I didn't (probably didn't)? The code says it shouldn't have, hence this fix.

Anyways, I had @w00tedness double check on his machine this time, but I'd like you guys to try it as well.

* Add optional parameter to the file helper writeFileSync to optional overwrite the existing file.
* Changed the writeFileSync param to take a single object since there are now two optional params.
* Applied logic from !855 to the versionedFile as well.

### DIFF
--- a/lib/preprocessCss.js
+++ b/lib/preprocessCss.js
@@ -223,7 +223,7 @@ module.exports = (app, callback) => {
       }
 
       // compare existing content to new content before generating
-      if (!versionData && versionData !== versionCode) {
+      if (!versionData || versionData !== versionCode) {
         fsr.writeFileSync(versionFile, versionCode, ['📝', `${appName} writing new versioned CSS file to reflect new version ${app.get('appVersion')} to ${versionFile}`.green])
       }
     }
@@ -282,8 +282,9 @@ module.exports = (app, callback) => {
                 content = fs.readFileSync(outpath, 'utf8')
               }
 
-              // check existing file for matching content before writing
-              if (!content || content !== newCSS) {
+              // check that the newCSS has content before writing the file. It is possible that the less file has no valid CSS and results in an empty CSS file. In this case don't write it.
+              // also, check existing file for matching content before writing
+              if (newCSS !== '' && (!content || content !== newCSS)) {
                 fsr.writeFileSync(outpath, newCSS, ['📝', `${appName} writing new CSS file ${outpath}`.green])
               }
               resolve()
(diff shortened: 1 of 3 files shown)


alallier:
@w00tedness can you test this again.

alallier:
Looks good on my end now

w00tedness:
> @w00tedness can you test this again.

Pulled and tested. 
Works fine on fresh install and on `nodemon` restart when you save a `*.less` file

alallier:
Standby I have one more little tweak. Found another small little bug

lib/preprocessCss.js
```diff
@@ -223,7 +223,7 @@ module.exports = (app, callback) => {
       }
 
       // compare existing content to new content before generating
-      if (!versionData && versionData !== versionCode) {
+      if (!versionData || versionData !== versionCode) {
```
alallier, line 226: @Autre31415 does this looks okay, I didn't really play around with the version file at all

Autre31415: Yeah this looks right. It matches the rest of it now

lib/preprocessCss.js
```diff
@@ -282,8 +282,9 @@ module.exports = (app, callback) => {
                 content = fs.readFileSync(outpath, 'utf8')
               }
 
-              // check existing file for matching content before writing
-              if (!content || content !== newCSS) {
+              // check that the newCSS has content before writing the file. It is possible that the less file has no valid CSS and results in an empty CSS file. In this case don't write it.
```
Autre31415, line 285: Are we sure we don't want to print the empty file? An empty file seems like a pretty good indication something is wrong with the CSS, as opposed to no file at all which one might assume is a Roosevelt problem.

alallier: Let me give you an example. Image a file like this `colors.less`

```less
@green: #00c000;
@yellow: #ffed29;
@red: #ff6969;
@cyan: #00ced1;
@white: #fff;
```

This will never generate a `.css` file. Instead when the less is processed these values are sprinkled in the respective `.css` files. If you don't check the newCss content for empty you will write an empty file.

Every time the CSS get's processed it will diff the _old_ CSS against the _new_ CSS and it will write the blank file every time. This will also generate a log, when there is no CSS changes in the first place.

Autre31415: Ah alright, that makes sense.

alallier:
@Autre31415 should be set

PR #973, opened 2020-10-29
Make symlink creation a separate step ran first in app startup

alallier:
Move symlinks creation to a separate step so they could be used safely without having an order of operations problem. This is first thing done on app startup so all symlinks are in place before trying to do any processing of any files.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Next version
 
-- Put your changes here...
+- Moved symlink creation to a separate step and made it the first thing that is done on app startup
 
 ## 0.19.3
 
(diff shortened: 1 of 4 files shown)


alallier:
Ci is failing for a non-obvious reason

alallier:
Sorry I forgot to edit the change log, I can do that now

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## Next version
 
 - Put your changes here...
```
kethinov, line 5: Delete this line. The one below removes the need for it.

kethinov (changes requested):

CHANGELOG.md
alallier: Done

kethinov (approved):

PR #1026, opened 2021-03-15
Remove upgrade-insecure-request from helmet config

### LINKED ISSUE #964: Strange https redirect bug in prod mode only
- Create sample app
- Start it in prod mode
- Instead of visiting http://localhost:43711, visit http://[your IP]:43711
- You will be redirected to the same URL, but with HTTPS, and the page will not load because of a SSL error.

hudson-pace:
The default config for helmet's content security policy has been altered to remove the upgrade-insecure-request parameter. Closes #964 .

### DIFF
--- a/README.md
+++ b/README.md
@@ -401,6 +401,8 @@ Resolves to:
   - To disable multipart forms entirely, set `formidable` to `false`.
 
 - `helmet`: Parameters to pass to the [helmet](https://github.com/helmetjs/helmet) module.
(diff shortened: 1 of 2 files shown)


Autre31415:
The updated method for solving this looks good to me. @kethinov  do you concur?

Autre31415 (approved):

kethinov (approved):

PR #1044, opened 2021-05-05
Update Roosevelt compatibility for Node 16

### LINKED ISSUE #1039: Node 16 support
- `nvm install 16`
- Remove node_modules directory.
- `npm i`
- Stuff breaks.

lannonbr:
- Migrated `node-sass` which used `LibSass` to `sass` which uses `dart-sass` as [LibSass was deprecated](https://sass-lang.com/blog/libsass-is-deprecated)
- Fixed some things for failing unit tests

Closes #1039.


lib/preprocessCss.js
```diff
@@ -128,6 +128,7 @@ module.exports = (app, callback) => {
 
             // enable inline source mapping in dev mode
             if (env === 'development') {
+              options.outFile = ''
```
lannonbr, line 131: This was added as in the Sass documentation for [`sourceMap`](https://sass-lang.com/documentation/js-api#sourcemap), it states

> If this option is true, the path is assumed to be the outFile option with .map added to the end. If it’s true and the outFile option isn’t passed, it has no effect.

LibSass didn't seem to actually assert this statement but dart-sass does. Not sure if this is the best solution or not as `outFile` technically is not used in `preprocessCss` after the Sass / SCSS file is parsed and compiled down to CSS.

lib/sourceParams.js
```diff
@@ -374,8 +374,8 @@ module.exports = (params, app, appSchema) => {
       for (const paramKey in paramSet) {
         const param = paramSet[paramKey]
 
-        // recurse if param value is an object
-        if (param === Object(param)) {
+        // recurse if param value is an object (but not a function)
+        if (param === Object(param) && typeof param !== 'function') {
```
lannonbr, line 378: Added this change as it hit a function like `onServerStart` in the params and tried parsing it which caused an infinite loop and threw an error for max callstack reached.

lib/preprocessCss.js
kethinov, line 131:
```suggestion
              options.outFile = '' // This was added as in the Sass documentation for sourceMap, it states "If this option is true, the path is assumed to be the outFile option with .map added to the end. If it’s true and the outFile option isn’t passed, it has no effect."
```

kethinov (changes requested):

lib/preprocessCss.js
lannonbr: Added the comment into the file upon your suggestion.

kethinov:
Excellent, thanks for this.

kethinov (approved):

PR #1221, opened 2023-01-18
Added methods for allowlist and blocklist to accept wildcards

Nicholas-R-Payne:


kethinov (changes requested):
some improvements needed.

assuming a standard sample app:

works:

```
"exposeAll": true,
"blocklist": ["layouts/*"],
```

^ successfully excludes layouts/main.html

doesn't work:

```
"exposeAll": true,
"blocklist": ["layouts*"],
```

^ the above should exclude layouts/main.html but does not

also broken:

```
"exposeAll": true,
"blocklist": ["4*"],
```

^ should exclude 404.html but instead crashes

allowlist also has problems:

```
  "exposeAll": false,
  "blocklist": [],
  "allowlist": {"views.js": ["layouts/*"]},
```

writes this to views.js: `module.exports = {"layouts/":"<!DOCTYPE html><html lang=en><head>[...]`

it shouldn't be "layouts/" — it should provide the whole filename for each file

and:

```
  "exposeAll": false,
  "blocklist": [],
  "allowlist": {"views.js": ["4*"]},
```

should only allow 404.html but crashes

lib/viewsBundler.js
```diff
@@ -3,6 +3,7 @@ require('@colors/colors')
 // Views Bundler: Module to expose templates that are available on the server down to the client as well
 
 module.exports = app => {
+  const fs = require('fs')
```
kethinov, line 6: i believe `fse` defined on the next line allows you to do all the same stuff as `fs`, so this line is irrelevant. all the `fs` calls below can be done from `fse` with an identical api

kethinov (approved):

PR #1341, opened 2023-12-20
Add `express-session` dependency

awidener3:
Include `express-session` dependency in `setExpressConfigs.js`.


lib/setExpressConfigs.js
```diff
@@ -45,6 +46,14 @@ module.exports = function (app) {
   // set morgan
   app.set('morgan', morgan)
 
+  // enable express-session
+  app.use(session({
+    // warning: issues may occur if secret does not match what is used in `cookie-parser`
+    secret: 'secret key',
+    resave: false,
+    saveUninitialized: false
+  }))
```
kethinov, line 55: our config would need to be a default config with an option to override it if we do end up doing this

kethinov (changes requested):

kethinov (approved):

PR #1351, opened 2024-01-22
Implement CSRF protection

### LINKED ISSUE #27: CSRF support
Some general interesting info on the topic:

- https://scotthelme.co.uk/csrf-is-really-dead/
- And https://news.ycombinator.com/item?id=28054292

After due consideration, we've decided to adopt [malibu](https://github.com/tinyhttp/malibu) at the framework level.

Will also need to add [express-session](https://github.com/expressjs/session) as a dependency as well.

From malibu docs:

Import deps:

```javascript
import session from 'express-session'
import { csrf } from 'malibu'
```

Load the middleware:

```javascript
const csrfProtection = csrf({ middleware: 'session' })
app.use(session({ secret: 'secret key', resave: false, saveUninitialized: false }))
```

Note `app` will already be defined by roosevelt.

Then each route will need to receive the `csrfProtection` variable. We'll need to document examples in the Roosevelt docs. Here's some examples from malibu:

```javascript
// this lets you acquire CSRF token on response body
app.get('/', csrfProtection, (req, res) => {
  res.status(200).json({ token: req.csrfToken() })
})

// you may only access this if you give a previously acquired CSRF token
app.post('/', csrfProtection, (req, res) => {
  res.status(200).json({ message: 'hello' })
})
```

I'm assuming that if a route omits `csrfProtection` then  it simply won't populate `req.csrfToken` and that isn't necessary to modify every route declaration to include the middleware unless the user decides they want CSRF protection on that route. That assumption needs to be tested.

### LINKED ISSUE #1244: Upgrade dev_sync.sh to a smarter Node.js-based script
I feel like we should upgrade this to a script written in Node that does intelligent things like prompt you for missing info and has more defensive code / error handling / etc in general.

awidener3:
Include CSRF protection in Roosevelt applications
- CSRF protection is enabled by default.
- CSRF can only be fully disabled by disabling `express-session` (added with #1341).
- Developers can switch between `full` (default) and `manual`. `manual` allows the dev to specify which routes are protected.
- The README has been updated to educate devs on how to include a CSRF token in their `POST` requests.

closes #27 

closes #1244


awidener3:
@kethinov I'm not 100% sold on my decision of using `full` vs. `manual` as configuration options. Any suggestions or thoughts?

README.md
```diff
@@ -490,8 +489,15 @@ Resolves to:
 
   - If `expressSession` is set to `true`, it will use the configuration below:
 
+    - Roosevelt uses [memorystore](https://github.com/roccomuso/memorystore) as the default `express-session` store.
+
+    - To use the default store provided by `express-session`, use `"store": "memorystore"`
```
kethinov, line 494: change this from `"store": "memorystore"` to `"store": "express-session-default"`

lib/setExpressConfigs.js
```diff
@@ -49,24 +51,75 @@ module.exports = function (app) {
   // enable express-session
   if (params.expressSession) {
     const secret = JSON.parse(fs.readFileSync(params.secretsDir + '/sessionSecret.json', 'utf-8')).secret
+    let store
+
+    if (params.store === 'default') {
+      if (process.env.NODE_ENV === 'production') {
```
kethinov, line 57: let's remove the mode check and do it in all modes

kethinov (changes requested):

PR #1457, opened 2024-08-28
Added the ability to exclude frontend reload script injection via configuration

### LINKED ISSUE #1150: Add exception routes configuration to reload module
Similar to #1149, when setting up frameworks that spin up their own dynamic routes we can end up in situations where the reload script tag gets appended to things that it shouldn't. A quick solution for niche scenarios like this is to allow users to configure a route or prefix of routes where the reload script tag won't get appended to a http response object.

alallier:
Added the ability to exclude frontend reload script injection via glob patterns at the server configuration level.

* Added minimatch as a dependency
  * I opted to do this as it's a npm level dependency and it reduces confusing custom REGEX checking (happy to change)
* Added new tests to test reload excludeRoutes
  * Added some more testing routes to correctly test the glob excluding in the reload configuration
* Added default excludeRoutes as empty array

(Closes #1150)


test/frontendReload.js
```diff
@@ -163,6 +163,35 @@ describe('frontend reload', () => {
...
+      frontendReload: {
+        ...config.frontendReload,
+        excludeRoutes: [
+          '/HTMLTest/*',
+          '/HTMLTest3'
+        ]
+      }
+    })
+
+    // check that reload script is injected into response bodies
```
alallier, line 180: Copy/paste comment is no longer correct

test/util/mvc/controllers/plainHTMLController.js
```diff
@@ -7,4 +7,18 @@ module.exports = (router) => {
     // send the plain HTML page back
     res.sendFile(htmlPath)
   })
+
+  router.route('/HTMLTest/nested').get((req, res) => {
+    // save the path of the plain HTML page
+    const htmlPath = path.join(__dirname, '../views/plainHTMLTest.html')
```
alallier, line 13: To make these more re-useable later and semantically correct we might want new .html documents later.

Let me know and I can update

alallier:
We need to hold merge on this for a bit, the test was working before but now it's breaking and there are recursion depth problems that could cause performance issues. Need to handle both of those items before merging

README.md
```diff
@@ -372,10 +372,13 @@ Resolves to:
     {
       "enable": true,
       "port": 9856,
-      "httpsPort": 9857
+      "httpsPort": 9857,
+      "excludeRoutes": []
     }
     ```
 
+  - `excludeRoutes`: Reload auto script injection happens on all HTML routes by default unless specified in `excludeRoutes`. An array of valid glob patterns is accepted
```
kethinov, line 380:
```suggestion
  - `excludeRoutes`: List of routes to exclude from Reload automatically injecting its script onto.
    - Default: *[Array]* with no items in it. This means Reload will inject its script on all routes by default.
```

kethinov (changes requested):

kethinov:
Screenshot 2024-08-28 at 8 49 52 PM

git push force

README.md
```diff
@@ -372,10 +372,13 @@ Resolves to:
     {
       "enable": true,
       "port": 9856,
-      "httpsPort": 9857
+      "httpsPort": 9857,
+      "excludeRoutes": []
```
Autre31415, line 376: The HTML minifier has a similar param called `exceptionRoutes`, either this new one, or the old one should be renamed for consistency.

kethinov:
superseded by https://github.com/rooseveltframework/roosevelt/releases/tag/0.26.0

